PR #7, opened 2016-02-26
Convert PRJEB/ERZ IDs also when returning VariantSource objects

cyenyxe:
The conversion was properly done in the web services that return Variant objects, but not in those that return VariantSource objects.

### DIFF
--- a/eva-lib/src/main/java/uk/ac/ebi/variation/eva/lib/storage/metadata/VariantSourceEvaproDBAdaptor.java
+++ b/eva-lib/src/main/java/uk/ac/ebi/variation/eva/lib/storage/metadata/VariantSourceEvaproDBAdaptor.java
@@ -131,7 +131,7 @@ public class VariantSourceEvaproDBAdaptor implements VariantSourceDBAdaptor {
         try {
             conn = ds.getConnection();
             pstmt = conn.prepareStatement(
-                    "select distinct f.ftp_file " +
+                    "select distinct bf.filename, f.ftp_file " +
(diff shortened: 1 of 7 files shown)


eva-lib/src/main/java/uk/ac/ebi/variation/eva/lib/storage/metadata/VariantSourceEvaproDBAdaptor.java
```diff
@@ -142,12 +142,12 @@ public QueryResult getSourceDownloadUrlByName(String filename) {
...
             if (rs.next()) {
-                URL url = new URL("ftp:/" + rs.getString(1));
+                URL url = new URL("ftp:/" + rs.getString(2));
                 long end = System.currentTimeMillis();
-                qr = new QueryResult(null, ((Long) (end - start)).intValue(), 1, 1, null, null, Arrays.asList(url));
+                qr = new QueryResult(rs.getString(1), ((Long) (end - start)).intValue(), 1, 1, null, null, Arrays.asList(url));
             } else {
                 long end = System.currentTimeMillis();
-                qr = new QueryResult(null, ((Long) (end - start)).intValue(), 0, 0, null, null, new ArrayList<>());
+                qr = new QueryResult(rs.getString(1), ((Long) (end - start)).intValue(), 0, 0, null, null, new ArrayList<>());
```
pabarcgar: I this will fail, if rs.next() returns false, rs.getString() could launch an SQLException

PR #13, opened 2016-09-06
EVA-440 Removed conflicting logging dependencies

cyenyxe:
Spring Boot favours Logback as logging library, that can be easily integrated with SLF4J. It was not possible to deploy the WAR file due to an extra SLF4J declaration.


jorizci:
I've tested with a clean docker-tomcat machine, reproduced the error and tested after applying the patch, the application is deployed without issues.

PR #15, opened 2016-09-22
EVA-451 allow to specify the read preference for mongo

jmmut:
also add tests for DBAdaptorConnector

test methods:
- correct autowiring for MultiMongoDbFactory
- "read preference" option properly loaded from properties

### DIFF
--- a/eva-lib/src/main/java/uk/ac/ebi/variation/eva/lib/datastore/DBAdaptorConnector.java
+++ b/eva-lib/src/main/java/uk/ac/ebi/variation/eva/lib/datastore/DBAdaptorConnector.java
@@ -22,6 +22,7 @@ package uk.ac.ebi.variation.eva.lib.datastore;
 import com.mongodb.*;
 import org.opencb.datastore.core.config.DataStoreServerAddress;
 import org.opencb.opencga.lib.auth.IllegalOpenCGACredentialsException;
+import org.opencb.opencga.lib.common.*;
 import org.opencb.opencga.storage.core.adaptors.StudyDBAdaptor;
 import org.opencb.opencga.storage.core.variant.adaptors.VariantDBAdaptor;
 import org.opencb.opencga.storage.core.variant.adaptors.VariantSourceDBAdaptor;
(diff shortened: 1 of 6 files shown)


eva-lib/src/main/java/uk/ac/ebi/variation/eva/lib/datastore/DBAdaptorConnector.java
```diff
@@ -105,7 +104,11 @@ public static MongoClient getMongoClient(Properties properties) throws UnknownHo
                     properties.getProperty("eva.mongo.passwd").toCharArray()));
         }
 
-        return new MongoClient(servers, mongoCredentialList);
+        MongoClientOptions options = MongoClientOptions.builder()
+                .readPreference(ReadPreference.valueOf(properties.getProperty("eva.mongo.read-preference")))
```
cyenyxe: A reasonable default for this `getProperty` would be "secondaryPreferred". It follows systems' recommendation, and if there are no secondaries then it automatically sends the requests to the primary.

jmmut: fixed in next commit

eva-server/src/test/java/uk/ac/ebi/variation/eva/lib/datastore/DBAdaptorConnectorTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Check that the value:
+     * eva.mongo.read-preference=secondaryPreferred
+     * is properly read by DBAdaptorConnector::getMongoClient
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testDefaultReadPreferenceInMongoClient() throws Exception {
```
cyenyxe: This is not really testing the default now, since there shouldn't be any value set in the properties file in order to do that. In addition, it would be good to have a method that checks wrong paths, using manually created properties with some other fields missing.

jmmut: fixed in next commit

PR #18, opened 2016-10-12
EVA-480 Migrate EVAPRO query layer to Spring and/or JPA

jorizci:
A few notes:

Database table: study_browser

study_type is a char varying that should mirror VariantStudy.StudyType, but we can find cases in the database like: "Case-Control" that appear as "Case Control"

variant_count: Variant count is a BigInt -> Long in java. The table field accepts null terminated values but the end class VariantStudy uses a int value. The code originally used ResultSet.getInt("param") to get the value. This method return 0 in case of null values. In the case of num Samples, its int in the VariantStudy, Integer in the table and, although the column still accept null values there are no null values.

The class models could be swapped with its entities counterparts.


eva-lib/src/main/java/uk/ac/ebi/eva/lib/spring/data/entity/DgvaStudyBrowser.java
```diff
@@ -0,0 +1,256 @@
...
+    private String platformName;
+
+    @Column(name = "assembly_name")
+    private String assemblyName;
+
+    public String getStudyAccession() {
+        return studyAccession;
+    }
+
+    public void setStudyAccession(String studyAccession) {
```
cyenyxe: Yes please, these objects are only read from the database, never modified.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/spring/data/entity/File.java
```diff
@@ -0,0 +1,176 @@
...
+    )
+})
+@Table(name = "file")
+public class File {
+
+    @Id
+    @Column(name = "file_id")
+    private Long file_id;
+
+    @Column(length = 45, name="ena_submission_file_id")
```
jorizci: Yes, it's one of the few tables with well limited characteristics.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/spring/data/entity/Project.java
```diff
@@ -0,0 +1,256 @@
...
+    private String projectAccession;
+
+    @Column(length = 250, nullable = false, name= "center_name")
+    private String centerName;
+
+    @Column(length = 4000, nullable = false)
+    private String alias;
+
+    @Lob
+    private String title;
```
jorizci: Its defined as text with no limit.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/spring/data/entity/StudyBrowser.java
```diff
@@ -0,0 +1,280 @@
...
+    @Column(name = "project_accession")
+    private String projectAccession;
+
+    @Column(name = "studyId")
+    private long studyId;
+
+    @Column(name = "projectTitle")
+    private String projectTitle;
+
+    private String description;
```
jorizci: This comes from a view without any limitation, all strings are defined as TEXT technically I should define here everything as Lob I just refrained as practically none of the elements are really text and if I use lob they will be handled as big text with lazy loading by default.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/spring/data/entity/Taxonomy.java
```diff
@@ -0,0 +1,100 @@
...
+    @Column(name = "taxonomy_id")
+    private Long taxonomyId;
+
+    @Column(length = 45, name = "common_name")
+    private String commonName;
+
+    @Column(length = 45, name = "scientific_name")
+    private String scientificName;
+
+    @Column(length = 100, name = "taxonomy_code")
```
jorizci: 45 Is what the db says...

cyenyxe:
IIRC the `FileFtpReference` was to be placed inside the `models` package.

jorizci:
Moved to models

PR #19, opened 2016-10-20
EVA-471 Securing REST API

jorizci:
~~Added new property~~
~~`eva.server.security.enabled`~~

After enabled it can be used to check a Bearer oauth2 credential against a server using the standard configuration properties from Spring-oauth2.

`security.oauth2.resource.user-info-uri`

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Build
 
 In order to build EVA, you need to install the Java Development Kit 7 and Maven.
 
-The project dependencies are OpenCGA and Variation-Commons 
+The project dependencies are OpenCGA and Variation-Commons
 
 You can get OpenCGA 0.5.2 from https://github.com/opencb/opencga, branch `hotfix/0.5`. Please follow the download/compilation instructions there.
 
@@ -28,3 +28,11 @@ The tests implemented so far are integration (not unit) tests, so a working WAR 
 3. Run `mvn jetty:run` from the eva-server subfolder
 4. Run `mvn test` from the root folder
 
+Enabling Oauth2 Security
+------------------------
+
+In order to enable Oauth2 Security you must enable the `oauth2-security` profile in the spring-boot application in any
+way supported by Spring-boot as stated in [Spring documentation](http://docs.spring.io/spring-boot/docs/current/reference/html/howto-properties-and-configuration.html#howto-set-active-spring-profiles). In adition, the application requires configuration to point to the public user check endpoints
+of your oauth2 compatible authentication service. More information is available in [Spring documentation](http://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-security.html#boot-features-security-oauth2-resource-server).
+
+When using Oauth2 Security, swagger UI will be still enabled and public to review the API but no command sent from the ui will be authorized.
--- a/README.md
+++ b/README.md
@@ -28,11 +28,9 @@ The tests implemented so far are integration (not unit) tests, so a working WAR 
 3. Run `mvn jetty:run` from the eva-server subfolder
 4. Run `mvn test` from the root folder
 
-Enabling Oauth2 Security
+Enabling OAuth2 Security
 ------------------------
 
-In order to enable Oauth2 Security you must enable the `oauth2-security` profile in the spring-boot application in any
-way supported by Spring-boot as stated in [Spring documentation](http://docs.spring.io/spring-boot/docs/current/reference/html/howto-properties-and-configuration.html#howto-set-active-spring-profiles). In adition, the application requires configuration to point to the public user check endpoints
-of your oauth2 compatible authentication service. More information is available in [Spring documentation](http://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-security.html#boot-features-security-oauth2-resource-server).
+In order to enable OAuth2 Security you must enable the `oauth2-security` profile in the spring-boot application in any way supported by Spring-boot as stated in [Spring documentation](http://docs.spring.io/spring-boot/docs/current/reference/html/howto-properties-and-configuration.html#howto-set-active-spring-profiles). In adition, the application requires configuration to point to the public user check endpoints of your OAuth2 compatible authentication service. More information is available in [Spring documentation](http://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-security.html#boot-features-security-oauth2-resource-server).
 
-When using Oauth2 Security, swagger UI will be still enabled and public to review the API but no command sent from the ui will be authorized.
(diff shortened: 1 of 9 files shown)


eva-server/src/main/java/uk/ac/ebi/eva/server/security/authorization/Oauth2Configuration.java
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+import org.springframework.security.oauth2.config.annotation.web.configuration.EnableResourceServer;
+import org.springframework.security.oauth2.config.annotation.web.configuration.ResourceServerConfigurerAdapter;
+
+
+@ConditionalOnProperty(prefix = "eva.server.security", name = "enabled", havingValue = "true")
+@Configuration
+@EnableResourceServer
+public class Oauth2Configuration extends ResourceServerConfigurerAdapter {
```
cyenyxe: Could a bit of documentation for enabling the OAuth server (even just a link to Spring docs) be added to the README file?

eva-server/src/main/java/uk/ac/ebi/eva/server/security/authorization/UnsecureConfiguration.java
```diff
@@ -0,0 +1,23 @@
+package uk.ac.ebi.eva.server.security.authorization;
+
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+import org.springframework.security.oauth2.config.annotation.web.configuration.EnableResourceServer;
+import org.springframework.security.oauth2.config.annotation.web.configuration.ResourceServerConfigurerAdapter;
+
+@ConditionalOnProperty(prefix = "eva.server.security", name = "enabled", matchIfMissing = true, havingValue = "false")
```
cyenyxe: Is this preferable over injecting the configuration using profiles?

jorizci: Need to check the behaviour, profiles seemed to work only to inject or not different application property values and/or override settings.

cyenyxe: If I understand [this documentation](http://docs.spring.io/spring/docs/current/spring-framework-reference/html/beans.html#beans-definition-profiles-java) correctly, they also allow to configure which `@Configuration` beans to inject.

jorizci:
Added documentation to readme, modified the approach to use profiles and only a profile when enabling oauth2 security.

README.md
```diff
@@ -28,3 +28,11 @@ The tests implemented so far are integration (not unit) tests, so a working WAR
...
 4. Run `mvn test` from the root folder
 
+Enabling Oauth2 Security
+------------------------
+
+In order to enable Oauth2 Security you must enable the `oauth2-security` profile in the spring-boot application in any
+way supported by Spring-boot as stated in [Spring documentation](http://docs.spring.io/spring-boot/docs/current/reference/html/howto-properties-and-configuration.html#howto-set-active-spring-profiles). In adition, the application requires configuration to point to the public user check endpoints
+of your oauth2 compatible authentication service. More information is available in [Spring documentation](http://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-security.html#boot-features-security-oauth2-resource-server).
+
+When using Oauth2 Security, swagger UI will be still enabled and public to review the API but no command sent from the ui will be authorized.
```
cyenyxe: Is there a way at all for people to send their tokens during testing via Swagger?

cyenyxe: Is there a way at all for people to send their tokens during testing via Swagger?

eva-server/src/main/java/uk/ac/ebi/eva/server/Profiles.java
```diff
@@ -0,0 +1,9 @@
+package uk.ac.ebi.eva.server;
+
+/**
+ * Created by jorizci on 01/11/16.
+ */
+public class Profiles {
+    public static final String OAUTH_SECURITY = "oauth2-security";
+    public static final String NOT_OAUTHSECURITY = "!"+ OAUTH_SECURITY;
```
cyenyxe: Right now we only have OAuth and non-OAuth, but in the future there will be more authentication schemas. Could this be renamed to something more user-friendly like "no-security" or "open"?

README.md
```diff
@@ -28,3 +28,11 @@ The tests implemented so far are integration (not unit) tests, so a working WAR
 3. Run `mvn jetty:run` from the eva-server subfolder
 4. Run `mvn test` from the root folder
 
+Enabling Oauth2 Security
+------------------------
+
+In order to enable Oauth2 Security you must enable the `oauth2-security` profile in the spring-boot application in any
+way supported by Spring-boot as stated in [Spring documentation](http://docs.spring.io/spring-boot/docs/current/reference/html/howto-properties-and-configuration.html#howto-set-active-spring-profiles). In adition, the application requires configuration to point to the public user check endpoints
```
cyenyxe: Please let the lines go as long as they can instead of adding arbitrary line breaks.

README.md
```diff
@@ -28,3 +28,11 @@ The tests implemented so far are integration (not unit) tests, so a working WAR
...
+of your oauth2 compatible authentication service. More information is available in [Spring documentation](http://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-security.html#boot-features-security-oauth2-resource-server).
```
cyenyxe: Please keep the capitalization consistent, some places say "OAuth2" and other "oauth2", some say "UI" and others "ui".

eva-server/src/main/java/uk/ac/ebi/eva/server/Profiles.java
jorizci: Ok, a little explanation, NOT_OAUTHSECURITY is not a profile, it means oauth-security profile is not enabled, so the condition if we add more system will be NOT_SEC_A and NOT_SEC_B, etc... That makes that by default if you don't have to modify the configuration to use the unsecure configuration.

README.md
jorizci: It seems [not yet](https://github.com/swagger-api/swagger-ui/issues/807)...

Tecnically the only supported way now is... you can put a Oauth2 login before the swagger page and pass the appropriate headers but that ties it to a specific system.

jorizci: But then I've found [this](http://www.appsdev.is.ed.ac.uk/blog/?p=820)... I'm not sure of how it will go. I need to test fully the example.

PR #23, opened 2016-12-12
EVA-366 VariantWSServer using VariantEntityRepository

tomdcsmith:
- getVariantById in VariantWSServer now uses VariantEntityRepository to query the mongo database
- New Utils class to convert a relational operator string to the enum value from VariantEntityRepository
- Also in Utils, PageRequest generation


eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
```diff
@@ -47,38 +54,65 @@
...
+                                                    String siftScore,
                                         HttpServletResponse response)
             throws IllegalOpenCGACredentialsException, UnknownHostException, IOException {
         initializeQueryOptions();
 
-        VariantDBAdaptor variantMongoDbAdaptor = DBAdaptorConnector.getVariantDBAdaptor(species);
 
-        if (studies != null && !studies.isEmpty()) {
-            queryOptions.put("studies", studies);
+        if (species == null || species.isEmpty()) {
```
jorizci: Species will never be null.

eva-server/src/main/java/uk/ac/ebi/eva/server/Utils.java
```diff
@@ -0,0 +1,41 @@
+package uk.ac.ebi.eva.server;
```
cyenyxe: Copyright header please.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
```diff
@@ -47,38 +54,65 @@
...
+            response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+            return setQueryResponse("Please specify a species");
         }
 
-        if (!variantId.contains(":")) { // Query by accession id
-            return setQueryResponse(variantMongoDbAdaptor.getVariantById(variantId, queryOptions));
-        } else { // Query by chr:pos:ref:alt
```
cyenyxe: This branch of the condition also needs to be supported for the Variant View to work, which means another method will be necessary in the repository.

Even if the all code called it `getAllVariantsByRegion`, this is not a query by region and could probably be generated automatically by Spring Data.

tomdcsmith:
Tests for queries greater than 1 million should be added to both variant ws and region ws tests.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
```diff
@@ -47,40 +54,90 @@
...
+        }
 
-            return setQueryResponse(variantMongoDbAdaptor.getAllVariantsByRegion(region, queryOptions));
+
+        QueryResult<VariantEntity> queryResult = new QueryResult<>();
+        queryResult.setResult(variantEntities);
+        return setQueryResponse(queryResult);
+    }
+
+    List<VariantEntity> queryByCoordAndAllele(String variantId) {
```
cyenyxe: Please rename to `queryByCoordinatesAndAlleles` and ensure the interfaces takes the arguments the method name declares: instead of taking a single string as argument, implement another auxiliary method to split the 3-4 parts of the string.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
```diff
@@ -47,40 +54,90 @@
...
+            variantEntities = queryByCoordAndAllele(variantId);
+        } else {
+            variantEntities = querybyId(variantId, studies, consequenceType, maf, polyphenScore, siftScore);
+        }
 
-            return setQueryResponse(variantMongoDbAdaptor.getAllVariantsByRegion(region, queryOptions));
+
+        QueryResult<VariantEntity> queryResult = new QueryResult<>();
+        queryResult.setResult(variantEntities);
+        return setQueryResponse(queryResult);
```
cyenyxe: The `QueryResult` object doesn't contain values for `numResults` or `numTotalResults`, which will break the pagination on the website. This affects the query by region more since that one will respond with multiple pages most of the times.

tomdcsmith: Are numResults the number of results that are being returned in the current query (current page), and the numTotalResults are the total number of results that match the query?

cyenyxe: Yes to both.

tomdcsmith:
- In new changes there is a new class that processes the filter values/relations, and holds their values (VariantFilterValues)
- A few new helper methods that are used in both the find and count queries
- the number of results and total number of the query result object is set

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
```diff
@@ -47,37 +54,86 @@
...
+    List<VariantEntity> queryByCoordinatesAndAlleles(String chromosome, int start, String reference, String alternate) {
+        if (alternate != null) {
+            return variantEntityRepository.findByChromosomeAndStartAndReferenceAndAlternate(chromosome, start,
+                                                                                            reference, alternate);
+        } else {
+            return variantEntityRepository.findByChromosomeAndStartAndReference(chromosome, start, reference);
+        }
+    }
+
+    Long countByCoordinatesAndAlleles(String chromosome, int start, String reference, String alternate) {
```
cyenyxe: This method should be private.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
```diff
@@ -47,37 +54,86 @@
...
 
-            return setQueryResponse(variantMongoDbAdaptor.getAllVariantsByRegion(region, queryOptions));
+        QueryResult<VariantEntity> queryResult = new QueryResult<>();
+        queryResult.setNumResults(variantEntities.size());
+        queryResult.setNumTotalResults(numTotalResults);
+        queryResult.setResult(variantEntities);
+        return setQueryResponse(queryResult);
+    }
+
+    List<VariantEntity> queryByCoordinatesAndAlleles(String chromosome, int start, String reference, String alternate) {
```
cyenyxe: This method should be private.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantFilterValues.java
```diff
@@ -0,0 +1,85 @@
+package uk.ac.ebi.eva.server.ws;
+
+import uk.ac.ebi.eva.lib.repository.VariantEntityRepository;
+import uk.ac.ebi.eva.server.Utils;
+
+class VariantFilterValues {
```
cyenyxe: The setter methods in this class should be private, because attribute values are set at creation time and never modified. A filter could be set atomically, as in:

```java
private void setMafFilter(String relation, Double value) {
    setMafOperator(Utils.getRelationalOperatorFromRelation(relation));
    setMafvalue(Utils.getValueFromRelation(value));
}
```

This small change could be beneficial for the refactoring to a generic solution for any filter.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantFilterValues.java
```diff
@@ -5,81 +5,62 @@
 
 class VariantFilterValues {
 
-    private VariantEntityRepository.RelationalOperator mafOperator;
+    private VariantEntityRepository.RelationalOperator mafOperator = VariantEntityRepository.RelationalOperator.NONE;
```
cyenyxe: This default value assignment is missing from the other 2 relational operators.

tomdcsmith:
Added tests for new count methods.

PR #24, opened 2016-12-12
EVA-366 RegionWSServer using VariantEntityRepository

tomdcsmith:
- Regression Test for RegionWSServer
- RegionWSServer now uses VariantEntityRepository to query the MongoDB database

### DIFF
--- a/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
+++ b/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
@@ -66,9 +66,9 @@ public class RegionWSServer extends EvaWSServer {
                                              @RequestParam(name = "species") String species,
                                              @RequestParam(name = "studies", required = false) List<String> studies,
                                              @RequestParam(name = "annot-ct", required = false) List<String> consequenceType,
-                                             @RequestParam(name = "maf", defaultValue = "", required = false) String maf,
-                                             @RequestParam(name = "polyphen", defaultValue = "", required = false) String polyphenScore,
-                                             @RequestParam(name = "sift", defaultValue = "", required = false) String siftScore,
+                                             @RequestParam(name = "maf", required = false) String maf,
+                                             @RequestParam(name = "polyphen", required = false) String polyphenScore,
+                                             @RequestParam(name = "sift", required = false) String siftScore,
                                              HttpServletResponse response)
             throws IllegalOpenCGACredentialsException, IOException {
         initializeQueryOptions();
@@ -82,7 +82,7 @@ public class RegionWSServer extends EvaWSServer {
 
         VariantEntityRepository.RelationalOperator mafOperator = VariantEntityRepository.RelationalOperator.NONE;
         Double mafvalue = null;
-        if (maf != null && !maf.isEmpty()) {
+        if (maf != null) {
             mafOperator = Utils.getRelationalOperatorFromRelation(maf);
             mafvalue = Utils.getValueFromRelation(maf);
         }
@@ -90,14 +90,14 @@ public class RegionWSServer extends EvaWSServer {
         VariantEntityRepository.RelationalOperator polyphenScoreOperator =
                 VariantEntityRepository.RelationalOperator.NONE;
         Double polyphenScoreValue = null;
-        if (polyphenScore != null && !polyphenScore.isEmpty()) {
+        if (polyphenScore != null) {
             polyphenScoreOperator = Utils.getRelationalOperatorFromRelation(polyphenScore);
             polyphenScoreValue = Utils.getValueFromRelation(polyphenScore);
         }
 
         VariantEntityRepository.RelationalOperator siftScoreOperator = VariantEntityRepository.RelationalOperator.NONE;
         Double siftScoreValue = null;
-        if (siftScore != null && !siftScore.isEmpty()) {
+        if (siftScore != null) {
             siftScoreOperator = Utils.getRelationalOperatorFromRelation(siftScore);
             siftScoreValue = Utils.getValueFromRelation(siftScore);
         }
--- a/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
+++ b/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
@@ -65,9 +65,9 @@ public class RegionWSServer extends EvaWSServer {
                                              @RequestParam(name = "species") String species,
                                              @RequestParam(name = "studies", required = false) List<String> studies,
                                              @RequestParam(name = "annot-ct", required = false) List<String> consequenceType,
-                                             @RequestParam(name = "maf", defaultValue = "") String maf,
-                                             @RequestParam(name = "polyphen", defaultValue = "") String polyphenScore,
-                                             @RequestParam(name = "sift", defaultValue = "") String siftScore,
+                                             @RequestParam(name = "maf", defaultValue = "", required = false) String maf,
+                                             @RequestParam(name = "polyphen", defaultValue = "", required = false) String polyphenScore,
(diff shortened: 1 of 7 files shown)


eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
```diff
@@ -55,98 +64,70 @@ public RegionWSServer() {
...
-                                             @RequestParam(name = "alt", defaultValue = "") String alternate,
-                                             @RequestParam(name = "miss_alleles", defaultValue = "") String missingAlleles,
-                                             @RequestParam(name = "miss_gts", defaultValue = "") String missingGenotypes,
-                                             @RequestParam(name = "histogram", defaultValue = "false") boolean histogram,
-                                             @RequestParam(name = "histogram_interval", defaultValue = "-1") int interval,
-                                             @RequestParam(name = "merge", defaultValue = "false") boolean merge,
+                                             @RequestParam(name = "maf", defaultValue = "", required = false) String maf,
+                                             @RequestParam(name = "polyphen", defaultValue = "", required = false) String polyphenScore,
+                                             @RequestParam(name = "sift", defaultValue = "", required = false) String siftScore,
```
jorizci: If defaultValue is present, the parameter value will always be that value when not present, If defaultValue is not defined, the value returned will be null by design. With that in mind, the sanity checks can be simplified.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
```diff
@@ -55,98 +64,70 @@ public RegionWSServer() {
...
                                              HttpServletResponse response)
             throws IllegalOpenCGACredentialsException, IOException {
         initializeQueryOptions();
 
-        VariantDBAdaptor variantMongoDbAdaptor = DBAdaptorConnector.getVariantDBAdaptor(species);
+        if (species == null || species.isEmpty()) {
+            response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+            return setQueryResponse("Please specify a species");
+        }
```
jorizci: Part of this can be passed to define studies as required and Spring will generate a BadRequest and an automated message of the problem in the parsing. 

This kind of checking/return is usually done better by raising an exception that can create the appropriate error response in Spring. I don't know if the scope of maintaining current API applies to this or not, @cyenyxe ?

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
```diff
@@ -55,98 +64,70 @@ public RegionWSServer() {
 //    @ApiOperation(httpMethod = "GET", value = "Retrieves all the variants from region", response = QueryResponse.class)
     public QueryResponse getVariantsByRegion(@PathVariable("regionId") String regionId,
                                              @RequestParam(name = "species") String species,
-                                             @RequestParam(name = "studies", required = false) String studies,
+                                             @RequestParam(name = "studies", required = false) List<String> studies,
```
jorizci: Doesn't this break API compatibility?

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
```diff
@@ -55,98 +64,70 @@ public RegionWSServer() {
...
+                    variantEntityRepository.findByRegionAndComplexFilters(region.getChromosome(), region.getStart(),
+                                                                          region.getEnd(), studies, consequenceType,
+                                                                          mafOperator, mafvalue, polyphenScoreOperator,
+                                                                          polyphenScoreValue, siftScoreOperator,
+                                                                          siftScoreValue, pageRequest);
         } else {
-            response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
-            return setQueryResponse("The total size of all regions provided can't exceed 1 million positions. "
-                    + "If you want to browse a larger number of positions, please provide the parameter 'histogram=true'");
+            throw new IllegalArgumentException();
```
jorizci: So, are we following the throw exceptions that are translated or the setQueryResponse? @cyenyxe

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
```diff
@@ -55,98 +64,70 @@ public RegionWSServer() {
...
+                                                                           polyphenScoreValue, siftScoreOperator,
+                                                                           siftScoreValue, pageRequest);
+        } else if (regions.size() == 1) {
+            Region region = regions.get(0);
+            variantEntities =
+                    variantEntityRepository.findByRegionAndComplexFilters(region.getChromosome(), region.getStart(),
+                                                                          region.getEnd(), studies, consequenceType,
+                                                                          mafOperator, mafvalue, polyphenScoreOperator,
+                                                                          polyphenScoreValue, siftScoreOperator,
+                                                                          siftScoreValue, pageRequest);
```
jorizci: I don't think the complexity of maintaining two methods that can generate equivalent queries is worth the marginal performance gain, if there is no good reason to have two methods in the repository, one that checks for the chromosome and other for a list of regions we should unite them. That way we can do a query with only one region in the list and simplify the code.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
tomdcsmith: [The VariantWSServer accepts a list of studies](https://github.com/EBIvariation/eva-ws/blob/develop/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java#L53), so I changed this to be consistent. @cyenyxe will this change cause any problems?

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
tomdcsmith: For sake of safety- in case the default values are removed, for example- could leaving the sanity checks as they are make sense?

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
jorizci: Then the check null is not needed, and my question still stands as how to proceed with error and if we should throw a exceptions as you do at line 125.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
jorizci: Call me old fashioned, but I prefer a RAII approach, check only one in initialization, otherwise we tend to fill the code with checkings. Also, another way to do it even better would be add JavaEE standard validation in the parameters and let spring handle the rest... But that may be much work right now and can be done later.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
cyenyxe: As far as I know Spring is able to split a string in comma-separated values so nothing should break, would be more an extension.

And as @tomdcsmith said, the other endpoint supports it and this is the only way to provide a consistent API for the Variant Browser.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
jorizci: Then it's better to do not throw exceptions right now as it's done on line 125?

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
cyenyxe: `setQueryResponse` for now, sorry guys.

PR #25, opened 2017-01-13
EVA-612 Option to exclude fields from documents returned

tomdcsmith:
- Added option to webservices to specify fields to exclude from being returned
- Modified repository methods to exclude the specified fields from being returned by Mongo

### DIFF
--- a/eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryTest.java
+++ b/eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryTest.java
@@ -27,8 +27,6 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.opencb.biodata.models.feature.Region;
 import org.opencb.biodata.models.variant.VariantSourceEntry;
-import org.opencb.biodata.models.variant.annotation.VariantAnnotation;
-import org.opencb.biodata.models.variant.stats.VariantStats;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.Bean;
@@ -52,7 +50,6 @@ import uk.ac.ebi.eva.commons.models.metadata.VariantEntity;
 
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -61,9 +58,7 @@ import java.util.Set;
 import static com.lordofthejars.nosqlunit.mongodb.MongoDbRule.MongoDbRuleBuilder.newMongoDbRule;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
 /**
@@ -84,6 +79,35 @@ public class VariantEntityRepositoryTest {
 
     @Autowired
     private VariantEntityRepository variantEntityRepository;
+
+    @Test
+    public void checkFieldPresence() throws IOException {
+
+        List<Region> regions = new ArrayList<>();
+        regions.add(new Region("11", 183000, 183300));
+
+        List<String> exclude = new ArrayList<>();
+
+        List<VariantEntity> variantEntityList = variantEntityRepository
+                .findByRegionsAndComplexFilters(regions, null, null,
+                                                VariantEntityRepository.RelationalOperator.NONE,
+                                                null,
+                                                VariantEntityRepository.RelationalOperator.NONE,
+                                                null,
+                                                VariantEntityRepository.RelationalOperator.NONE,
+                                                null, exclude,
+                                                new PageRequest(0, 100000000));
+
+        for (VariantEntity currVariantEntity : variantEntityList) {
+            assertFalse(currVariantEntity.getSourceEntries().isEmpty());
+            assertFalse(currVariantEntity.getIds().isEmpty());
+            for (Map.Entry<String, VariantSourceEntry> sourceEntryEntry :
+                    currVariantEntity.getSourceEntries().entrySet()){
+                assertFalse(sourceEntryEntry.getValue().getAttributes().isEmpty());
+            }
+        }
+
+    }
 
     @Test
     public void testVariantIdIsFound(){
@@ -474,6 +498,8 @@ public class VariantEntityRepositoryTest {
         assertEquals(expectedResultLength, variantEntityList.size());
     }
 
+    // config class
+    // @import
     @Configuration
     @EnableMongoRepositories
     @ComponentScan(basePackageClasses = { VariantEntityRepository.class })
--- a/eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryTest.java
+++ b/eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryTest.java
@@ -311,7 +311,6 @@ public class VariantEntityRepositoryTest {
 
     @Test
     public void testFindByRegionsAndComplexFiltersExclude() {
-
         List<Region> regions = new ArrayList<>();
         regions.add(new Region("11", 183000, 183300));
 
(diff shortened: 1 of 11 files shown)


eva-server/src/main/java/uk/ac/ebi/eva/server/Utils.java
```diff
@@ -59,4 +75,19 @@ public static PageRequest getPageRequest(QueryOptions queryOptions) {
...
+            excludeArray[i] = String.format("'%s' : 0", excludeList.get(i));
+        }
+        return "{ " + StringUtils.join(excludeArray, ", ") + " }";
+    }
+
+    private static void initApiToMongoDocNameMap() {
+        apiToMongoDocNameMap = new HashMap<>();
+        apiToMongoDocNameMap.put("sourceEntries", "files");
+        apiToMongoDocNameMap.put("sourceEntries.statistics", "st");
+        apiToMongoDocNameMap.put("annotation.statistics", "annot");
```
cyenyxe: Just "annotation", without nested "statistics".

eva-server/src/main/java/uk/ac/ebi/eva/server/Utils.java
```diff
@@ -18,13 +18,29 @@
...
+
 public class Utils {
 
+    private static Map<String, String> apiToMongoDocNameMap;
+
+    static {
+        initApiToMongoDocNameMap();
+    }
+
+    public static Map<String, String> getApiToMongoDocNameMap() {
```
cyenyxe: Instead of this, a method like the following could be used to hide the fact that we use a Map and avoid the need to return an unmodifiable copy.

```java
public static String getDatabaseFieldMapping(String apiFieldName) {
    return apiToMongoDocNameMap.get(apiFieldName);
}
```

It would also be easier to test (see comment on `UtilsTest` class).

eva-server/src/main/java/uk/ac/ebi/eva/server/Utils.java
```diff
@@ -18,13 +18,29 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
 public class Utils {
 
+    private static Map<String, String> apiToMongoDocNameMap;
+
+    static {
+        initApiToMongoDocNameMap();
```
cyenyxe: The intention could be absolutely clarified by making the method return a value and assign it to the static variable:

```java
static {
    apiToMongoDocNameMap = initApiToMongoDocNameMap();
}
```

eva-server/src/test/java/uk/ac/ebi/eva/server/UtilsTest.java
```diff
@@ -0,0 +1,24 @@
+package uk.ac.ebi.eva.server;
```
cyenyxe: Add licensing header (dated 2017).

eva-server/src/test/java/uk/ac/ebi/eva/server/UtilsTest.java
```diff
@@ -0,0 +1,24 @@
...
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.Assert.*;
+
+/**
+ * @author Tom Smith
+ */
```
cyenyxe: Remove author information.

eva-server/src/test/java/uk/ac/ebi/eva/server/UtilsTest.java
```diff
@@ -0,0 +1,24 @@
...
+public class UtilsTest {
```
cyenyxe: If the API to Mongo mapping is extracted to `getDatabaseFieldMapping`, then 4 new test methods can be added here to check that all 3 fields are correctly mapped, and the rest do not have a mapping.

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryTest.java
```diff
@@ -272,37 +287,81 @@ public void testFindByRegionsAndComplexFilters() {
...
         regions.add(new Region("11", 180205, 180221)); //2
 
         testFindByRegionsAndComplexFiltersHelper(regions, null, null, VariantEntityRepository.RelationalOperator.NONE,
                                                  null, VariantEntityRepository.RelationalOperator.NONE, null,
-                                                 VariantEntityRepository.RelationalOperator.NONE, null, 11);
+                                                 VariantEntityRepository.RelationalOperator.NONE, null, null, 11);
+    }
+
+    @Test
+    public void testFindByRegionsAndComplexFiltersExclude() {
```
tomdcsmith: Single and multiple field tests have been added, and run ok, but nested field exclusion aren't currently supported by the Fongo version that is used (https://github.com/fakemongo/fongo/commit/3f937e16dbc11604b3e9e97ef9e0a1ae9c598911- and the version that supports nested exclusion in Fongo does not support version 3 of the Mongo driver)

eva-server/src/main/java/uk/ac/ebi/eva/server/Utils.java
```diff
@@ -59,4 +78,19 @@ public static PageRequest getPageRequest(QueryOptions queryOptions) {
...
+                                                .collect(Collectors.toList());
+        return "{ " + String.join(", ", formattedList) + " }";
+    }
+
+    private static Map<String, String> initApiToMongoDocNameMap() {
+        Map<String, String> map = new HashMap<>();
+        map.put("sourceEntries", "files");
+        map.put("sourceEntries.statistics", "st");
+        map.put("annotation", "annot");
+        map.put("filesAttrs", "files.attrs");
```
cyenyxe: The key should be `sourceEntries.attributes` to match the names returned by the web service.

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryTest.java
```diff
@@ -272,37 +287,163 @@ public void testFindByRegionsAndComplexFilters() {
...
         regions.add(new Region("11", 180205, 180221)); //2
 
         testFindByRegionsAndComplexFiltersHelper(regions, null, null, VariantEntityRepository.RelationalOperator.NONE,
                                                  null, VariantEntityRepository.RelationalOperator.NONE, null,
-                                                 VariantEntityRepository.RelationalOperator.NONE, null, 11);
+                                                 VariantEntityRepository.RelationalOperator.NONE, null, null, 11);
+    }
+
+    @Test
+    public void testFindByRegionsAndComplexFiltersExcludeSingleRoot() {
```
cyenyxe: This method is testing multiple test cases: excluding a one-level subdocument, and not excluding any fields. Therefore it should be split in 2.

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryTest.java
```diff
@@ -272,37 +287,163 @@ public void testFindByRegionsAndComplexFilters() {
...
+                                                                       null, exclude,
+                                                                       new PageRequest(0, 100000000));
+
+        for (VariantEntity currVariantEntity : variantEntityList) {
+            assertFalse(currVariantEntity.getSourceEntries().isEmpty());
+        }
+    }
+
+    @Test
+    public void testFindByRegionsAndComplexFiltersExcludeMultiple() {
```
cyenyxe: One of the test cases (no exclusions) will be redundant if the previous one is split in 2. The asserts can be grouped under that single method.

Applies also to `testFindByRegionsAndComplexFiltersExcludeNested`.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
```diff
@@ -85,6 +87,22 @@ public QueryResponse getVariantsByRegion(@PathVariable("regionId") String region
         List<Region> regions = Region.parseRegions(regionId);
 
         PageRequest pageRequest = Utils.getPageRequest(queryOptions);
+        logger.info("Got page request");
```
cyenyxe: The logging messages in this class should be debug level, because they allow low level tracing of the application.

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/RegionWSServerTest.java
```diff
@@ -51,6 +52,19 @@ public void testGetVariantsByRegions() throws URISyntaxException {
         testGetVariantsByRegionHelper("20:60000-61000,20:61500-62500", 17);
     }
 
+    @Test
+    public void testExcludeNested() throws URISyntaxException {
+        Response response = given().param("species", "mmusculus_grcm38").param("exclude", "filesAttrs").get(new URI("/v1/segments/20:60000-62000/variants"));
```
cyenyxe: The key should be `sourceEntries.attributes` to match the names returned by the web service.

PR #26, opened 2017-01-17
EVA-612 Moved Mongo driver version to version 3

tomdcsmith:
- Added exclusions for mongo driver to dependencies
- Mongo 3 dependency from Spring Boot dependency
- Initialised credentials list, cannot be null in v3 of the driver- was preventing successful deployment

### DIFF
--- a/eva-lib/pom.xml
+++ b/eva-lib/pom.xml
@@ -23,6 +23,10 @@
             </exclusions>
         </dependency>
         <dependency>
+            <groupId>uk.ac.ebi.eva</groupId>
+            <artifactId>variation-commons</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.postgresql</groupId>
             <artifactId>postgresql</artifactId>
         </dependency>
--- a/eva-lib/pom.xml
+++ b/eva-lib/pom.xml
@@ -23,11 +23,6 @@
             </exclusions>
         </dependency>
         <dependency>
-            <groupId>uk.ac.ebi.eva</groupId>
-            <artifactId>variation-commons</artifactId>
-            <version>0.2-SNAPSHOT</version>
-        </dependency>
-        <dependency>
             <groupId>org.postgresql</groupId>
             <artifactId>postgresql</artifactId>
         </dependency>
--- a/eva-lib/pom.xml
+++ b/eva-lib/pom.xml
@@ -34,7 +34,6 @@
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-data-mongodb</artifactId>
-            <version>1.4.2.RELEASE</version>
         </dependency>
         <dependency>
             <groupId>org.springframework.boot</groupId>
--- a/eva-lib/pom.xml
+++ b/eva-lib/pom.xml
@@ -15,15 +15,40 @@
         <dependency>
             <groupId>org.opencb.opencga</groupId>
             <artifactId>opencga-storage-mongodb</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.mongodb</groupId>
+                    <artifactId>mongo-java-driver</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>uk.ac.ebi.eva</groupId>
             <artifactId>variation-commons</artifactId>
             <version>0.2-SNAPSHOT</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.springframework.boot</groupId>
+                    <artifactId>spring-boot-starter-data-jpa</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.springframework.boot</groupId>
+                    <artifactId>spring-boot-starter-data-mongodb</artifactId>
+                </exclusion>
+                <exclusion>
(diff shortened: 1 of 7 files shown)


eva-lib/pom.xml
```diff
@@ -15,24 +15,49 @@
...
+                <exclusion>
+                    <groupId>org.mongodb</groupId>
+                    <artifactId>mongo-java-driver</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>uk.ac.ebi.eva</groupId>
             <artifactId>variation-commons</artifactId>
             <version>0.2-SNAPSHOT</version>
```
cyenyxe: Given that `variation-commons` is an internal dependency, it would be best to upgrade the Spring Boot starters to the versions that we want to use, instead of excluding them.

The OpenCGA exclusion is fine because we will remove that dependency at some point.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/MongoConfiguration.java
```diff
@@ -16,6 +16,7 @@
 package uk.ac.ebi.eva.lib;
 
 import com.mongodb.MongoClient;
+import com.mongodb.MongoClientURI;
```
cyenyxe: This import is not used.

eva-server/pom.xml
```diff
@@ -55,11 +55,35 @@
...
+                    <groupId>org.springframework.boot</groupId>
+                    <artifactId>spring-boot-starter-data-mongodb</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.mongodb</groupId>
+                    <artifactId>mongo-java-driver</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.opencb.opencga</groupId>
+                    <artifactId>opencga-storage-mongodb</artifactId>
```
cyenyxe: Where does `variation-commons` have an OpenCGA dependency?

eva-server/pom.xml
```diff
@@ -55,11 +55,35 @@
...
+                <exclusion>
+                    <groupId>org.mongodb</groupId>
+                    <artifactId>mongo-java-driver</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>uk.ac.ebi.eva</groupId>
             <artifactId>variation-commons</artifactId>
             <version>0.2-SNAPSHOT</version>
```
cyenyxe: Same comment as for the `eva-lib` POM file.

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerTest.java
```diff
@@ -105,12 +105,12 @@ private void testGetVariantByIdRegionDoesntExistHelper(String testString) throws
 
     @Test
     public void testCheckVariantExistsDoesExist() throws URISyntaxException {
-        assertTrue(testCheckVariantExistsHelper("20:71822:C:G"));
+        testCheckVariantExistsHelper("20:71822:C:G");
```
cyenyxe: Why are these asserts removed?

eva-server/pom.xml
tomdcsmith: https://github.com/EBIvariation/variation-commons/blob/develop/pom.xml#L70 here

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerTest.java
tomdcsmith: I removed them because I thought they were failing for a different reason than they are, I thought they were failing because there wasn't meant to be an assert there, and that all the asserts happen in the helper.

eva-server/pom.xml
cyenyxe: Is there any class apart from [MongoDBObjectToVariantEntityConverter](https://github.com/EBIvariation/variation-commons/blob/develop/src/main/java/uk/ac/ebi/eva/commons/models/converters/data/MongoDBObjectToVariantEntityConverter.java) using OpenCGA code?

tomdcsmith: That's the only class that uses OpenCGA.

tomdcsmith:
Exclusions from variation-commons dependencies have been removed.

eva-lib/pom.xml
```diff
@@ -27,12 +33,17 @@
         </dependency>
         <dependency>
             <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-data-mongodb</artifactId>
+            <version>1.4.2.RELEASE</version>
```
cyenyxe: Version already specified in the parent POM.

pom.xml
```diff
@@ -24,7 +24,7 @@
     <properties>
         <opencga.version>0.5.2</opencga.version>
         <compileSource>1.8</compileSource>
-        <mongo-java-driver.version>2.14.2</mongo-java-driver.version>
+        <mongo-java-driver.version>3.2.2</mongo-java-driver.version>
```
cyenyxe: Version number no longer necessary, it is inherited from `spring-boot-starter-data-mongodb`.

PR #29, opened 2017-02-06
EVA-637 Improvement of EVA-WS Spring configuration

jorizci:
Moved and refactored repository configuration.

### DIFF
--- a/eva-server/src/main/java/uk/ac/ebi/eva/server/configuration/MongoEvaLibConfiguration.java
+++ b/eva-server/src/main/java/uk/ac/ebi/eva/server/configuration/MongoEvaLibConfiguration.java
@@ -1,0 +1,12 @@
+package uk.ac.ebi.eva.server.configuration;
+
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+import org.springframework.data.mongodb.repository.config.EnableMongoRepositories;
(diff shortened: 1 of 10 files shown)


eva-server/src/main/java/uk/ac/ebi/eva/server/configuration/MongoEvaLibConfiguration.java
```diff
@@ -0,0 +1,12 @@
...
+
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+import org.springframework.data.mongodb.repository.config.EnableMongoRepositories;
+import uk.ac.ebi.eva.lib.MongoConfiguration;
+
+@Configuration
+@EnableMongoRepositories("uk.ac.ebi.eva.lib.repository")
+@Import(MongoConfiguration.class)
+public class MongoEvaLibConfiguration {
```
jmmut: I don't understand the purpose of this class, isn't it the same as MongoConfiguration?

jorizci: Not exactly, It says two things include the values in MongoEvaLibConfiguration and then enable mongo repositories. MongoConfiguration doesn't configure the repositories in eva-lib.

PR #31, opened 2017-02-09
EVA-613 Study web services using Spring Data (2/2)

jmmut:
this pull request requires https://github.com/EBIvariation/variation-commons/pull/23 to be merged first

this code replaces the "studies/{study}/view" and "archive/studies/list" requests with spring data repositories.

It's possible that this PR needs the dependency version upgrade made in master. I'll take care of this if travis doesn't pass after merging  https://github.com/EBIvariation/variation-commons/pull/23.


eva-lib/src/test/java/uk/ac/ebi/eva/lib/configuration/MongoRepositoryTestConfiguration.java
```diff
@@ -19,63 +19,42 @@
...
-    @Autowired
-    private MongoDbFactory mongoDbFactory;
-
-    @Override
-    protected String getDatabaseName() {
-        return "test-db";
-    }
+@Import({MongoConfiguration.class})
+@PropertySource({"classpath:eva.properties"})
+public class MongoRepositoryTestConfiguration {
```
cyenyxe: What is the reason for making this not extend `AbstractMongoConfiguration` any more?

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantStudySummaryRepositoryTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+        next = studiesIterator.next();
+        assertEquals(FIRST_STUDY_ID, next.getStudyId());
+        assertEquals(FIRST_STUDY_NAME, next.getStudyName());
+        assertCorrectCount(EXPECTED_FILE_COUNT_FROM_FIRST_STUDY_ID, next);
+        next = studiesIterator.next();
+        assertEquals(SECOND_STUDY_ID, next.getStudyId());
+        assertEquals(SECOND_STUDY_NAME, next.getStudyName());
+        assertCorrectCount(EXPECTED_FILE_COUNT_FROM_SECOND_STUDY_ID, next);
+
```
cyenyxe: Remove blank line.

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantStudySummaryRepositoryTest.java
```diff
@@ -0,0 +1,125 @@
...
+import static com.lordofthejars.nosqlunit.mongodb.MongoDbRule.MongoDbRuleBuilder.newMongoDbRule;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(classes = {MongoRepositoryTestConfiguration.class})
+@UsingDataSet(locations = {"/test-data/files.json"})
+public class VariantStudySummaryRepositoryTest {
```
cyenyxe: Please add tests for non existing studies, we need to check both positive and negative behavior. Rename test methods accordingly.

When there is not a single condition to check, prefix test methods with `test` to follow naming conventions (e.g. `testListStudies`).

eva-lib/src/test/resources/eva.properties
```diff
@@ -6,7 +6,7 @@ eva.mongo.auth.db=testdb
 eva.mongo.read-preference=
 
 eva.mongo.collections.variants=vars
-eva.mongo.collections.files=files
+eva.mongo.collections.files=files_1_2
```
cyenyxe: It should be possible to make this work using something a different name from "files_1_2", right? If so, it would be good to keep it different to avoid any confusion.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/ArchiveWSServer.java
```diff
@@ -110,8 +115,14 @@ public QueryResponse getStudies(@RequestParam(name = "species", required = false
     @RequestMapping(value = "/studies/list", method = RequestMethod.GET)
     public QueryResponse getBrowsableStudies(@RequestParam("species") String species)
             throws IllegalOpenCGACredentialsException, IOException {
-        StudyDBAdaptor studyMongoDbAdaptor = DBAdaptorConnector.getStudyDBAdaptor(species);
-        return setQueryResponse(studyMongoDbAdaptor.listStudies());
+        MultiMongoDbFactory.setDatabaseNameForCurrentThread(DBAdaptorConnector.getDBName(species));
+        List<VariantStudySummary> uniqueStudies = variantStudySummaryRepository.findBy();
+
+        QueryResult<VariantStudySummary> result = new QueryResult<>();
```
cyenyxe: With these usages in addition to those in [RegionWSServer](https://github.com/EBIvariation/eva-ws/blob/develop/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java) this `QueryResult` initialization is getting very redundant. Could this small code block be extracted to a method in the parent class?

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantStudySummaryRepositoryImpl.java
```diff
@@ -0,0 +1,130 @@
...
+    private static final String ID = "_id.";
+
+    private MongoTemplate mongoTemplate;
+
+    public VariantStudySummaryRepositoryImpl(MongoDbFactory mongoDbFactory, MappingMongoConverter mappingMongoConverter) {
+        mongoTemplate = new MongoTemplate(mongoDbFactory, mappingMongoConverter);
+    }
+
+    /**
+     * the equivalent intended query is:
```
jmmut: I put "the intended" because if there were a mismatch between both queries, the one in the comment is the intended one. If there were a mismatch without knowing which one is the correct, it might lead to "improve" the good one to a wrong query.

eva-lib/src/test/java/uk/ac/ebi/eva/lib/configuration/MongoRepositoryTestConfiguration.java
jmmut: The previous version of this class was mixing too many things, and overlapped with MongoConfiguration. Once we removed redundancy across other classes, the `AbstractMongoConfiguration` was just a constraint on what methods to override, and was not really needed.

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantStudySummaryRepositoryTest.java
```diff
@@ -0,0 +1,131 @@
...
+import static com.lordofthejars.nosqlunit.mongodb.MongoDbRule.MongoDbRuleBuilder.newMongoDbRule;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(classes = {MongoRepositoryTestConfiguration.class})
+@UsingDataSet(locations = {"/test-data/files.json"})
```
tomdcsmith: @jmmut @cyenyxe do we want to use the same test data for the files and for the study repositories? If not then the name of the files.json test data file could be changed to reflect that it's just for testing the study repository.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantStudySummaryRepositoryImpl.java
```diff
@@ -0,0 +1,141 @@
...
+        );
+
+        AggregationResults<VariantStudySummary> studies = mongoTemplate.aggregate(aggregation,
+                VariantSourceEntity.class,
+                VariantStudySummary.class);
+
+        VariantStudySummary variantStudySummary;
+        if (studies.getMappedResults().size() == 0) {
+            variantStudySummary = null;
+        } else {
```
tomdcsmith: Is it possible that multiple results are returned?

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/StudyWSServer.java
```diff
@@ -85,22 +94,18 @@ public QueryResponse getStudy(@PathVariable("study") String study,
...
         initializeQueryOptions();
 
-        StudyDBAdaptor studyMongoDbAdaptor = DBAdaptorConnector.getStudyDBAdaptor(species);
+        MultiMongoDbFactory.setDatabaseNameForCurrentThread(DBAdaptorConnector.getDBName(species));
+        VariantStudySummary variantStudySummary = variantStudySummaryRepository.findByStudyNameOrStudyId(study);
 
-        QueryResult idQueryResult = studyMongoDbAdaptor.findStudyNameOrStudyId(study, queryOptions);
-        if (idQueryResult.getNumResults() == 0) {
-            QueryResult queryResult = new QueryResult();
+        QueryResult<VariantStudySummary> queryResult = new QueryResult<>();
```
tomdcsmith: `new QueryResult<>()` is redundant here

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
```diff
@@ -118,10 +118,8 @@ public QueryResponse getVariantById(@PathVariable("variantId") String variantId,
         }
 
 
-        QueryResult<VariantEntity> queryResult = new QueryResult<>();
-        queryResult.setNumResults(variantEntities.size());
+        QueryResult<VariantEntity> queryResult = Utils.buildQueryResult(variantEntities);
         queryResult.setNumTotalResults(numTotalResults);
```
tomdcsmith: To save having to do this in some of the classes, the Utils method buildQueryResult could be overloaded. e.g.:

    public static <T> QueryResult<T> buildQueryResult(List<T> results) {
        return buildQueryResult(results, results.size());
    }

and

    public static <T> QueryResult<T> buildQueryResult(List<T> results, int numTotalResults) {
        QueryResult<T> queryResult = new QueryResult<>();
        queryResult.setResult(results);
        queryResult.setNumResults(results.size());
        queryResult.setNumTotalResults(numTotalResults);
        return queryResult;
    }

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantStudySummaryRepositoryImpl.java
jmmut: if a document in the files collection has the same studyId as another document, but has a different study name, yes, it will happen. This would be a DB inconsistency and I don't really know what should we do here. any thoughts @cyenyxe ?

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantStudySummaryRepositoryTest.java
jmmut: If it is enough for the `file repository test` I would leave it as is. If the test needs more features in a sample collection to be tested properly, make one for each test and rename them

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantStudySummaryRepositoryImpl.java
cyenyxe: That should be prevented at loading time by defining the corresponding unique indexes, which we don't have right now because [this one](https://github.com/EBIvariation/eva-pipeline/blob/develop/src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VariantSourceEntityMongoWriter.java#L52) only checks file uniqueness.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/projections/VariantStudySummary.java
```diff
@@ -0,0 +1,68 @@
...
+
+    public int getFilesCount() {
+        return filesCount;
+    }
+
+    /**
+     * This method is required to use a Tree of VariantStudySummary to ignore duplicated summaries.
+     */
+    @Override
+    public int compareTo(Object o) {
```
cyenyxe: `equals`, `hashCode` and `compare` should use the same set of attributes or there would be multiple uniqueness criteria. If we use `compareTo` as reference, only the study ID should be used in all of them.

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantStudySummaryRepositoryTest.java
```diff
@@ -0,0 +1,127 @@
...
+    public void testFindsByNameOrIdProvidingName() {
+        assertFindBySecondNameOrId(SECOND_STUDY_NAME);
+    }
+
+    @Test
+    public void testFindsByNameOrIdProvidingId() {
+        assertFindBySecondNameOrId(SECOND_STUDY_ID);
+    }
+
+    public void assertFindBySecondNameOrId(String studyNameOrId) {
```
cyenyxe: private

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantStudySummaryRepositoryTest.java
```diff
@@ -0,0 +1,127 @@
...
+        assertEquals(FIRST_STUDY_ID, next.getStudyId());
+        assertEquals(FIRST_STUDY_NAME, next.getStudyName());
+        assertCorrectCount(EXPECTED_FILE_COUNT_FROM_FIRST_STUDY_ID, next);
+        next = studiesIterator.next();
+        assertEquals(SECOND_STUDY_ID, next.getStudyId());
+        assertEquals(SECOND_STUDY_NAME, next.getStudyName());
+        assertCorrectCount(EXPECTED_FILE_COUNT_FROM_SECOND_STUDY_ID, next);
+    }
+
+    public void assertCorrectCount(int expectedFileCount, VariantStudySummary study) {
```
cyenyxe: private

eva-server/src/main/resources/application.properties
```diff
@@ -17,4 +17,6 @@ spring.jmx.enabled = false
 springfox.documentation.swagger.v2.path=/webservices/api
 
 #spring.profiles.active=oauth2-security
-#security.oauth2.resource.user-info-uri = ...
\ No newline at end of file
+#security.oauth2.resource.user-info-uri = ...
+
+eva.mongo.collections.files=@eva.mongo.collections.files@
```
cyenyxe: I assume this parameter could be removed from `eva.properties` after all the WS that reference it are migrated?

jmmut: mmm, migrated to spring repositories? I don't think it could or should be removed, this is the place where the repositories will look for the collection names.

cyenyxe: I mean from the other properties file (`eva.properties` instead of `application.properties`).

jmmut: then yes. I think some work will have to be done on MultiMongoFactoryConfiguration (because it currently also reads from eva.properties), but I don't see big reasons against that.

PR #34, opened 2017-02-15
EVA-588 Evapro db tests

pabarcgar:
Added Evapro y Dgva adaptors tests

### DIFF
--- a/eva-lib/pom.xml
+++ b/eva-lib/pom.xml
@@ -57,6 +57,23 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.hsqldb</groupId>
+            <artifactId>hsqldb</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+            <version>2.8.6</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>2.8.6</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
     
 </project>
--- a/eva-lib/pom.xml
+++ b/eva-lib/pom.xml
@@ -60,7 +60,6 @@
         <dependency>
             <groupId>com.h2database</groupId>
             <artifactId>h2</artifactId>
-            <version>1.4.193</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/eva-lib/pom.xml
+++ b/eva-lib/pom.xml
@@ -58,8 +58,9 @@
             </exclusions>
         </dependency>
         <dependency>
-            <groupId>org.hsqldb</groupId>
-            <artifactId>hsqldb</artifactId>
+            <groupId>com.h2database</groupId>
+            <artifactId>h2</artifactId>
+            <version>1.4.193</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/eva-lib/pom.xml
+++ b/eva-lib/pom.xml
@@ -62,18 +62,6 @@
             <artifactId>h2</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-core</artifactId>
-            <version>2.8.6</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-databind</artifactId>
-            <version>2.8.6</version>
-            <scope>test</scope>
-        </dependency>
     </dependencies>
     
 </project>
--- a/eva-lib/src/test/java/uk/ac/ebi/eva/lib/metadata/ArchiveDgvaDBAdaptorTest.java
+++ b/eva-lib/src/test/java/uk/ac/ebi/eva/lib/metadata/ArchiveDgvaDBAdaptorTest.java
@@ -29,8 +29,7 @@ public class ArchiveDgvaDBAdaptorTest {
 
     @Before
     public void setUp() throws Exception {
-        DgvaStudyTestData testData = new DgvaStudyTestData(entityManager);
-        testData.persistTestData();
+        DgvaStudyTestData.persistTestData(entityManager);
     }
 
     @Test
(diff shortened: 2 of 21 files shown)


eva-lib/src/test/java/uk/ac/ebi/eva/lib/metadata/MetadataTestConfiguration.java
```diff
@@ -0,0 +1,15 @@
...
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
+
+import uk.ac.ebi.eva.lib.extension.ExtendedJpaRepositoryFunctionsImpl;
+
+@SpringBootConfiguration
+@ComponentScan
+@EntityScan(basePackages = {"uk.ac.ebi.eva.lib.entity"})
+@EnableJpaRepositories(basePackages = {"uk.ac.ebi.eva.lib.repository"}, repositoryBaseClass = ExtendedJpaRepositoryFunctionsImpl.class)
+public class MetadataTestConfiguration {
```
tomdcsmith: Is this class used for anything?

pabarcgar: This class is used by Spring to initialise the repositories and entities used in the test (it will create a in memory database containing all the entities). 

Like this module is a library, there is no spring boot application annotation (that would do the same) and we need this class (removing it will make the tests not work).

eva-lib/src/test/java/uk/ac/ebi/eva/lib/metadata/EvaStudyBrowserTestData.java
```diff
@@ -0,0 +1,55 @@
...
+
+    static final String EXOME_SEQUENCING = "Exome sequencing";
+
+    static final String PROJECT_ID_1 = "PRJ0001";
+
+    EvaStudyBrowserTestData(TestEntityManager entityManager) {
+        this.entityManager = entityManager;
+    }
+
+    void persistTestData() {
```
pabarcgar: You are right, updating ...

pabarcgar:
Added license to all modified or new files

eva-lib/pom.xml
```diff
@@ -57,6 +57,24 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>com.h2database</groupId>
+            <artifactId>h2</artifactId>
+            <version>1.4.193</version>
```
cyenyxe: Include versions in parent POM when possible.

eva-lib/pom.xml
```diff
@@ -57,6 +57,24 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
```
cyenyxe: What are the Jackson artifacts required for?

eva-lib/src/test/java/uk/ac/ebi/eva/lib/metadata/ArchiveDgvaDBAdaptorTest.java
```diff
@@ -0,0 +1,137 @@
...
+import uk.ac.ebi.eva.lib.utils.QueryOptionsConstants;
+
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(SpringRunner.class)
+@DataJpaTest
+public class ArchiveDgvaDBAdaptorTest {
```
cyenyxe: Add tests for those cases where the species or study type does not exist in the database. Also applies to `ArchiveEvaproDBAdaptorTest`.

eva-lib/src/test/java/uk/ac/ebi/eva/lib/metadata/EvaStudyBrowserTestData.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+package uk.ac.ebi.eva.lib.metadata;
+
+import org.springframework.boot.test.autoconfigure.orm.jpa.TestEntityManager;
+
+import uk.ac.ebi.eva.lib.entity.EvaStudyBrowser;
+
+final class EvaStudyBrowserTestData {
+
+    static final String HUMAN = "human";
```
cyenyxe: It could be convenient to store the scientific names in constants too.

eva-lib/src/test/java/uk/ac/ebi/eva/lib/metadata/StudyEvaproDBAdaptorTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Before
+    public void setUp() throws Exception {
+        EvaStudyBrowserTestData.persistTestData(entityManager);
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        entityManager.clear();
+    }
```
cyenyxe: Isn't it enough to make these `@BeforeClass` and `@AfterClass` methods? The class being tested only has read operations, so the data shouldn't change between test cases.

I don't know how costly is to create an in-memory `EntityManager`, but if they are remotely similar to the regular ones this could help to speed up the tests execution time.

Applies to other test classes.

eva-lib/pom.xml
pabarcgar: For some spring test data internals. Jackson is being imported through other dependency, so we can remove those safely from the pom.

eva-lib/pom.xml
pabarcgar: Done

eva-lib/src/test/java/uk/ac/ebi/eva/lib/metadata/StudyEvaproDBAdaptorTest.java
pabarcgar: The autowired "TestEntityManager" doesn't work in static enviroment, so we cannot use BeforeClass

eva-lib/src/test/java/uk/ac/ebi/eva/lib/metadata/ArchiveDgvaDBAdaptorTest.java
pabarcgar: Added in fb4e520

eva-lib/src/test/java/uk/ac/ebi/eva/lib/metadata/EvaStudyBrowserTestData.java
pabarcgar: done in 8b6640b

PR #37, opened 2017-03-13
EVA-681 GA4GH web services using Spring

tomdcsmith:
- New repository methods for querying variant entities and variant source entities
- GA4GH web services using Spring repositories


eva-server/src/main/java/uk/ac/ebi/eva/server/ws/ga4gh/GA4GHVariantWSServer.java
```diff
@@ -73,35 +83,40 @@ public GASearchVariantsResponse getVariantsByRegion(@RequestParam("referenceName
...
             queryOptions.put("skip", idxCurrentPage * limit);
         }
         queryOptions.put("limit", limit);
-        
-        // Create the provided region, whose size can't exceed 1 million positions
+
         Region region = new Region(chromosome, start, end);
-        int regionSize = region.getEnd()-region.getStart();
-        
-        if (regionSize > 1000000) {
```
tomdcsmith: Yeah, due to the new indexing and the method the VariantEntityRepositoryImpl queries the database.
@cyenyxe can confirm if I'm correct in thinking this.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantSourceEntityRepository.java
```diff
@@ -36,4 +39,14 @@
 
     List<VariantSourceEntity> findByStudyIdOrStudyName(String studyId, String studyName);
 
+    List<VariantSourceEntity> findByStudyIdIn(List<String> studyIds, Pageable pageable);
+
+    long countByStudyIdIn(List<String> studyIds);
+
+    @Query(value = "{fid: {$in : ?0}}}", fields = "{fid : 1, samp : 1}")
```
tomdcsmith: The reason I'm using this is because at the moment the only client code that uses this query method just requires the file ID and the sample name, thus I'm using a projection to only return these values.  
Do you think it would be better to remove the projection, in case future client code needs more of the document returning, to save it needing to be removed then?

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepository.java
```diff
@@ -60,4 +63,11 @@ Long countByChromosomeAndStartAndReferenceAndAlternate(String chromosome, int st
 
     @Query(value = "{'chr': ?0, 'start': ?1, 'ref': ?2}", count = true)
```
tomdcsmith: The queries don't work without the annotation. The reason is that it's querying for VariantEntity objects, but this class doesn't have document annotations for its variables since it's quite a basic subclass of the biodata variant object. The Mongo document is instead converted by the DBObjectToVariantEntityConverter class. 
The Variant class has different names for the alt and ref alleles than in the Mongo document ("alt" and "ref" in Mongo doc, "alternate" and "reference" in the Variant(Entity) class. So, for example, when a repository method is called "countByChromosomeAndStartAndRefAndAlt" Spring throws an error complaining:

"Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'variantEntityRepository': Invocation of init method failed; nested exception is org.springframework.data.mapping.PropertyReferenceException: No property ref found for type VariantEntity!"

and when the method is called "countByChromosomeAndStartAndReferenceAndAlternate" it just doesn't return anything. But it does work and return results with the Query annotation.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/ga4gh/GA4GHVariantWSServer.java
cyenyxe: Yes we can :D

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantSourceEntityRepository.java
jmmut: no, as general rule, the apis should be as tight as possible. this allows changes more easily, imagine that a user relies on a api feature that we provide because it's easy but not important. then we change arquitecture, and that field is still not important and difficult, so we break compatibility. If instead we force the user to use the main fields there might be some workaround, but no big deal with internal updates.

Ok, the reason to use the `@Query` was to do a projection. That's actually interesting, from the docs I thought the only way to do that was using another bean with a subset of fields

tomdcsmith: This is now removed

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepository.java
```diff
@@ -60,4 +63,11 @@ Long countByChromosomeAndStartAndReferenceAndAlternate(String chromosome, int st
...
     Long countByChromosomeAndStartAndReference(String chr, int start, String ref);
+
+    @Query(value = "{'chr': ?0, 'start': ?1, 'alt': ?2, 'files.sid': {$in : ?3}}}", count = true)
+    Long countByChromosomeAndStartAndEndAndAltAndStudyIn(String chr, int start, String alt, List<String> studyIds);
```
cyenyxe: `AndEnd` should not be present in the method name

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryImpl.java
```diff
@@ -81,7 +86,8 @@ public Long countByRegionsAndComplexFilters(List<Region> regions, List<VariantEn
         return countByComplexFiltersHelper(query, filters);
     }
 
-    private List<VariantEntity> findByComplexFiltersHelper(Query query, List<VariantEntityRepositoryFilter> filters, List<String> exclude,
+    private List<VariantEntity> findByComplexFiltersHelper(Query query, List<VariantEntityRepositoryFilter> filters,
+                                                           List<String> exclude,
                                                            Pageable pageable) {
```
cyenyxe: Line formatting: `pageable` should fit in the same line as `exclude`.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepository.java
```diff
@@ -60,4 +63,11 @@ Long countByChromosomeAndStartAndReferenceAndAlternate(String chromosome, int st
...
+
+    @Query(value = "{'chr': ?0, 'start': ?1, 'type': ?2, 'files.sid': {$in : ?3}}}", count = true)
```
cyenyxe: I wonder, can we use constants to define this string? That way the queries wouldn't need to be modified if the database schema changes.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/ga4gh/GA4GHVariantCallSetWSServer.java
```diff
@@ -72,24 +80,36 @@ public GASearchCallSetsResponse getCallSets(@RequestParam("variantSetIds") Strin
...
+                                                                .map(ArrayList::new)
+                                                                .collect(Collectors.toList());
+
         // Convert sample names objects to GACallSet
-        List<GACallSet> gaCallSets = GACallSetFactory.create(filesList, qr.getResult());
+        List<GACallSet> gaCallSets = GACallSetFactory.create(fileIds, samplesLists);
         // Calculate the next page token
         int idxLastElement = idxCurrentPage * limit + limit;
-        String nextPageToken = (idxLastElement < qr.getNumTotalResults()) ? String.valueOf(idxCurrentPage + 1) : null;
+        String nextPageToken = (idxLastElement < numTotalResults) ? String.valueOf(idxCurrentPage + 1) : null;
```
cyenyxe: I don't know how much the usage of `Pageable` (mentioned above) could also affect this code block, but anyway, could the `nextPageToken` initialization be encapsulated in a utility method as in [the modified EvaWSServer class](https://github.com/jmmut/eva/blob/60ffcf2a22e8ec178508fad5ad4ce9e5a680ec8b/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/EvaWSServer.java)?

Note that change will be merged in the tests PR, it is currently in the `Utils` class.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/ga4gh/GA4GHVariantCallSetWSServer.java
```diff
@@ -61,7 +69,7 @@ public GA4GHVariantCallSetWSServer() { }
      * @see http://ga4gh.org/documentation/api/v0.5/ga4gh_api.html#/schema/org.ga4gh.GASearchCallSetsRequest
      */
     @RequestMapping(value = "/search", method = RequestMethod.GET)
-    public GASearchCallSetsResponse getCallSets(@RequestParam("variantSetIds") String files,
+    public GASearchCallSetsResponse getCallSets(@RequestParam("variantSetIds") List<String> files,
                                                 @RequestParam(name = "pageToken", required = false) String pageToken,
                                                 @RequestParam(name = "pageSize", defaultValue = "10") int limit,
```
cyenyxe: GA4GH endpoints should use `Pageable` objects and remove the histogram functionality, in a similar fashion to the rest of the API.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepository.java
tomdcsmith: I'm not sure if this is possible, to somehow inject strings into query annotations. When we first started on the Spring repositories for variants, and we wanted to query start and end values where start was both less than the specified end, but also greater than the specified start minus a certain margin (e.g. `db.variants_1_2.find({ "chr" : 10 , "end"  : { $gt :  5000000, $lt : 6000000  }, "start"  : { $lt :  6000000, $gt : 4000000 }})` for a querying a variant at position 5000000).
I think this was one of the reasons the repository had to be manually implemented (in addition to needing to apply the filters).

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepository.java
```diff
@@ -60,4 +63,11 @@ Long countByChromosomeAndStartAndReferenceAndAlternate(String chromosome, int st
 
     @Query(value = "{'chr': ?0, 'start': ?1, 'ref': ?2}", count = true)
     Long countByChromosomeAndStartAndReference(String chr, int start, String ref);
+
+    @Query(value = "{'chr': ?0, 'start': ?1, 'alt': ?2, 'files.sid': {$in : ?3}}}", count = true)
+    Long countByChromosomeAndStartAndAltAndStudyIn(String chr, int start, String alt, List<String> studyIds);
+
+    @Query(value = "{'chr': ?0, 'start': ?1, 'type': ?2, 'files.sid': {$in : ?3}}}", count = true)
+    Long countByChromosomeAndStartAndTypeAndStudyIn(String chr, int start, Variant.VariantType type,
+                                                    List<String> studyIds);
```
cyenyxe: I am wondering about the usefulness of some of these `countBy` methods: for paginated queries they are useful because they allow to calculate the total number, not just for a single page. But for non-paginated results we could just have `findBy` methods that are be more flexible in terms of data that can be obtained. The number of results would be the size of the in-memory collection.

This would mean that:
* `countByChromosomeAndStartAndReference(AndAlternate)` are no longer necessary
* `countByChromosomeAndStartAndAltAndStudyIn` and `countByChromosomeAndStartAndTypeAndStudyIn` are converted into `findBy` methods

eva-lib/src/main/java/uk/ac/ebi/eva/lib/filter/Helpers.java
```diff
@@ -8,14 +10,26 @@
                                                                                         String polyphenScore,
                                                                                         String siftScore,
                                                                                         List<String> studies,
-                                                                                        List<String> consequenceType) {
+                                                                                        List<String> consequenceType,
+                                                                                        List<String> files,
+                                                                                        List<Variant.VariantType> types,
+                                                                                        List<String> alternates) {
```
cyenyxe: At the moment there are 2 very well delimited filter use cases:

* EVA endpoints: First 5 arguments
* GA4GH endpoints: "files" argument only

So it is better to create 2 methods, one for each case, than a completely generic one that is called with many null arguments.

Another option would be to use a builder object that would be totally flexible:

```java
List<VariantEntityRepositoryFilter> filters = withMaf(maf)
                                             .withPolyphenScore(polyphenScore)
                                             .withSiftScore(siftScore)
                                             .withStudies(studies);
```

This would be most convenient for testing as a lot of combinations are necessary, but not so much for the application code, so the suggestion about the 2 facade methods could still apply.

jmmut:
what's the state of this? we talked about adding tests for GA4GH like the ones added in #38, but are we going to include them in this PR? I think it's getting too big, I would favour merging (even if we merge to a branch different than develop), just to mark that this code is reviewed, and reduce the next review.

cyenyxe:
We can include the tests in another PR, but I would then leave the JIRA ticket open. In any case, there have been substantial changes after my latest review so I need to check those.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/filter/FilterBuilder.java
```diff
@@ -0,0 +1,84 @@
+package uk.ac.ebi.eva.lib.filter;
+
+import org.opencb.biodata.models.variant.Variant;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class FilterBuilder {
```
cyenyxe: License and class documentation.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/filter/FilterBuilder.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private List<VariantEntityRepositoryFilter> filters = new ArrayList<>();
+
+    public List<VariantEntityRepositoryFilter> getEvaWsVariantEntityRepositoryFilters(String maf,
```
cyenyxe: No need to specify "EvaWs", it's the repository we are already in.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/filter/FilterBuilder.java
```diff
@@ -0,0 +1,84 @@
...
+                   .withConsequenceType(consequenceType)
+                   .build();
+    }
+
+    public List<VariantEntityRepositoryFilter> build() {
+        return filters;
+    }
+
+    public FilterBuilder withMaf(String maf) {
+        if (maf != null){
```
cyenyxe: Please auto-format the code, there should be a whitespace after the parenthesis.

eva-server/src/main/java/uk/ac/ebi/eva/server/Utils.java
```diff
@@ -43,13 +48,28 @@
     public static PageRequest getPageRequest(QueryOptions queryOptions) {
         int limit = (int) queryOptions.get("limit");
         int skip = (int) queryOptions.get("skip");
```
cyenyxe: Set default values when the properties are undefined. Otherwise these casts will throw an exception. For consistency, they can be the same as in https://github.com/EBIvariation/eva-ws/pull/37/files#diff-45d2055383ecc49840bd0966d60042f9R62

eva-server/src/main/java/uk/ac/ebi/eva/server/Utils.java
```diff
@@ -65,4 +85,43 @@ public static String createExclusionFieldString(List<String> excludeList) {
...
+
+    public static <T> QueryResult<T> buildQueryResult(List<T> results, long numTotalResults) {
+        QueryResult<T> queryResult = new QueryResult<>();
+        queryResult.setResult(results);
+        queryResult.setNumResults(results.size());
+        queryResult.setNumTotalResults(numTotalResults);
+        return queryResult;
+    }
+
+    public static List<VariantSource> convertVariantSourceEntitysToVariantSources(List<VariantSourceEntity>
```
cyenyxe: "`VariantSourceEntit**ies**"

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
```diff
@@ -76,16 +76,16 @@ public QueryResponse getVariantsByRegion(@PathVariable("regionId") String region
             throws IllegalOpenCGACredentialsException, IOException {
         initializeQuery();
 
-        if (species.isEmpty()) {
+        if (species == null || species.isEmpty()) {
```
cyenyxe: This extra check is not necessary: species is required and will never be null.

PR #38, opened 2017-03-15
EVA-690 Improve testability of web services that depend on Mongo

jmmut:
things to note in this PR:
- DBAdaptorConnector had to be done non-static to be mocked
- the tests for the web services now operate on java types, instead of a JSON string
- neither rest-assured, jetty, a mongod instance or a tomcat instance are needed to run and pass the tests 
- moved buildQueryResult to EvaWSServer and added setErrorQueryResponse


eva-server/src/main/java/uk/ac/ebi/eva/server/ws/EvaWSServer.java
```diff
@@ -78,52 +82,72 @@ public Jackson2ObjectMapperBuilder jacksonBuilder() {
...
-        // Guarantee that the QueryResponse object contains a coll of results
-        List coll;
-        if (obj instanceof List) {
-            coll = (List) obj;
-        } else {
-            coll = new ArrayList();
-            coll.add(obj);
-        }
+
+        // TODO why is this lost? because it's a native int?
```
jmmut: the dbtime in the queryResponse object is not received in the tests, that's why I added it to the queryResult

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/FilesWSServerTest.java
```diff
@@ -18,64 +18,89 @@
...
-                assertTrue(missingField, stats.containsKey("passCount"));
-                assertTrue(missingField, stats.containsKey("transitionsCount"));
-                assertTrue(missingField, stats.containsKey("transversionsCount"));
-                assertTrue(missingField, stats.containsKey("accumulatedQuality"));
-                assertTrue(missingField, stats.containsKey("meanQuality"));
-            }
+        String url = "/v1/files/all?species=hsapiens_grch37";
+        ResponseEntity<QueryResponse<QueryResult<VariantSourceEntity>>> response = restTemplate.exchange(
+                url, HttpMethod.GET, null,
+                new ParameterizedTypeReference<QueryResponse<QueryResult<VariantSourceEntity>>>() {});
```
jmmut: in this version I was using getForObject, but changed it to remove casts and being sure what object was seralized: https://github.com/jmmut/eva/blob/5dc332b436f54cb3e0f5640ba9c61d7fd57308d4/eva-server/src/test/java/uk/ac/ebi/eva/server/ws/ArchiveWSServerTest.java

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/RegionWSServerTest.java
```diff
@@ -18,68 +18,90 @@
...
+
+        List<Region> twoRegions = (List<Region>) argThat(hasSize(2));
+        given(variantEntityRepository.findByRegionsAndComplexFilters(twoRegions, any(), any(), any()))
+                .willReturn(Arrays.asList(variantEntity, variantEntity));
     }
 
     @Test
     public void testGetVariantsByRegion() throws URISyntaxException {
-        testGetVariantsByRegionHelper("20:60000-62000", 40);
+        testGetVariantsByRegionHelper("20:60000-62000", 1);
```
jmmut: mockito. I told it in the setup that if the `variantEntityRepository.findByRegionsAndComplexFilters()` receives one region, it must return a list with one variant, and if the request has two regions, it returns two variants. the contents of the variants are never used. And I'm not sure we should encourage that confusion.

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryTest.java
```diff
@@ -339,6 +339,26 @@ public void testFindByRegionsAndComplexFiltersExcludeSingleRoot() {
         }
     }
 
+    @Test
+    public void testFindByRegionsAndComplexFiltersExcludeSourceEntries() {
+        List<Region> regions = new ArrayList<>();
+        regions.add(new Region("11", 183000, 183300));
+
+        List<String> exclude = new ArrayList<>();
+        exclude.add("files.attrs");
```
jmmut: Ok I see I made a mistake here. there's no "attrs" constant as changed in the previous one, because there's no filter by attrs. here we are using the "exclude" which has nothing to do with filters. I'll change this so that the exclude ones are retrieved from the converters instead.

pom.xml
```diff
@@ -140,6 +135,11 @@
                     </exclusion>
                 </exclusions>
             </dependency>
+            <dependency>
+                <groupId>org.mongodb</groupId>
+                <artifactId>mongo-java-driver</artifactId>
+                <version>${mongo-java-driver.version}</version>
+            </dependency>
```
jmmut: no, it's not necessary

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/FilesWSServerTest.java
cyenyxe: You are right, this is the most type-safe solution if the entities are not returned straight away.

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/RegionWSServerTest.java
cyenyxe: This way we are not testing what is going on in the web service body.

Imagine that a bug turns the chromosome into an empty string: since the mock repository always returns the same output, independently of the input provided, it would be impossible to detect said bug.

What you could do is to define a small set of regions:
* 20:60000-62000
* 20:61500-62500
* Anything else

And configure the repository to return values for the first 2, and nothing for anything else. That way we can also test the response for non-existing regions quite easily.

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryTest.java
```diff
@@ -339,6 +342,27 @@ public void testFindByRegionsAndComplexFiltersExcludeSingleRoot() {
         }
     }
 
+    @Test
+    public void testFindByRegionsAndComplexFiltersExcludeSourceEntries() {
```
jmmut: yes, test the exclusion of attributes as an example of nested attributes

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/FilesWSServerTest.java
```diff
@@ -18,64 +18,89 @@
...
+        assertEquals(HttpStatus.OK, response.getStatusCode());
+
+        QueryResponse<QueryResult<VariantSourceEntity>> queryResponse = response.getBody();
+        assertEquals(1, queryResponse.getResponse().size());
+
+        List<VariantSourceEntity> results = queryResponse.getResponse().get(0).getResult();
+        assertEquals(1, results.size());
+
+        for (VariantSourceEntity variantSourceEntity : results) {
```
pabarcgar: The for is not  really needed because is just one file

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/RegionWSServerTest.java
pabarcgar: +1

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/RegionWSServerTest.java
```diff
@@ -18,68 +18,105 @@
...
 package uk.ac.ebi.eva.server.ws;
 
-import com.jayway.restassured.RestAssured;
-import com.jayway.restassured.path.json.JsonPath;
-import com.jayway.restassured.response.Response;
-import org.junit.BeforeClass;
+import org.junit.Before;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.AdditionalMatchers;
```
pabarcgar: A couple of unused imports in that class

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/FilesWSServerTest.java
jmmut: yes, it's like that for similarity with other tests that do have more elements, and simplicity for extending this one.

 assigning to a simple variable did not provide much advantages, (get(0)? why only the first one?), and this highlights that every element should pass those checks

eva-lib/src/main/java/uk/ac/ebi/eva/lib/models/Assembly.java
```diff
@@ -37,8 +37,10 @@
     private String taxonomyCode;
     private String taxonomyEvaName;
 
-    
-    public Assembly(String assemblyAccession, String assemblyChain, String assemblyVersion, String assemblyName, String assemblyCode, 
+    Assembly() {
```
jmmut: "meta/species/list" calls `archiveEvaproDbAdaptor.getSpecies` which returns a `List<Assembly>`. the tests try to deserialize them from JSON to Java object. It is not strictly needed for the application, only for the tests.

PR #39, opened 2017-03-23
EVA-376 Query chromosomes from EVA

tomdcsmith:
Spring repository method to query Mongo collection for the distinct values for "chr".
Webservice utilising this repository for returning the list of chromosomes.


eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
```diff
@@ -114,4 +114,21 @@ public QueryResponse getVariantsByRegion(@PathVariable("regionId") String region
...
         return setQueryResponse("");
     }
+
+    @RequestMapping(value = "", method = RequestMethod.GET)
+    @ResponseBody
+//    @ApiOperation(httpMethod = "GET", value = "Retrieves all the variants from region", response = QueryResponse.class)
+    public QueryResponse getChromosomes(@RequestParam(name = "species") String species,
+                                        HttpServletResponse response)
+            throws IllegalOpenCGACredentialsException, IOException {
+        if (species == null || species.isEmpty()) {
```
cyenyxe: The null check should not be necessary.

cyenyxe: The null check should be unnecessary.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryImpl.java
```diff
@@ -81,6 +81,12 @@ public Long countByRegionsAndComplexFilters(List<Region> regions, List<VariantEn
         return countByComplexFiltersHelper(query, filters);
     }
 
+    @Override
+    public List<String> findDistinctChromosomes() {
+        return (List<String>) mongoTemplate.getCollection(mongoTemplate.getCollectionName(VariantEntity.class))
+                                           .distinct("chr", new Query().getQueryObject());
```
cyenyxe: It shouldn't be necessary to create an object for an empty query (at least the low level API allows to indicate only the distinct field).

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryTest.java
```diff
@@ -339,6 +339,16 @@ public void testFindByRegionsAndComplexFiltersExcludeSingleRoot() {
...
     }
 
+    @Test
+    public void testFindDistinctChromosomesByStudyId() {
+        List<String> chromosomeList = variantEntityRepository.findDistinctChromosomes();
+
+        List<String> expectedChromosomeList = new ArrayList<>();
+        expectedChromosomeList.add("11");
+
+        assertEquals(expectedChromosomeList, chromosomeList);
```
pabarcgar: expectedChromsomeList can be declared using `Collections.singletonList("11")`, inside the assertEquals (or outside if you prefer for readibility).

This is an small method and is not really necessary here, but in bigger test methods reducing the number of lines improves the readibility.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
```diff
@@ -114,4 +114,21 @@ public QueryResponse getVariantsByRegion(@PathVariable("regionId") String region
     public QueryResponse getVariantsByRegion() {
         return setQueryResponse("");
     }
+
+    @RequestMapping(value = "", method = RequestMethod.GET)
```
pabarcgar: @cyenyxe , Is the "segments" path clear enough, or should we add "/chromosomes"?

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
```diff
@@ -114,4 +114,21 @@ public QueryResponse getVariantsByRegion(@PathVariable("regionId") String region
...
+    @ResponseBody
+//    @ApiOperation(httpMethod = "GET", value = "Retrieves all the variants from region", response = QueryResponse.class)
+    public QueryResponse getChromosomes(@RequestParam(name = "species") String species,
+                                        HttpServletResponse response)
```
pabarcgar: In the vcf-dumper WS, study(ies) is a mandatory param (and files for the CLI). If the query speed is not affected by adding those filters, it would improve the vcf-dumper performance (for studies that don't have variants in each chromosome). If the speed of this query is affected, maybe is better not executing those filters, so we get the list of chrosomes and start exporting faster (this would be good for the WS, when a long waiting time since we call the WS until we get some data can confuse the user).

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryTest.java
```diff
@@ -339,6 +339,16 @@ public void testFindByRegionsAndComplexFiltersExcludeSingleRoot() {
         }
     }
 
+    @Test
+    public void testFindDistinctChromosomesByStudyId() {
+        List<String> chromosomeList = variantEntityRepository.findDistinctChromosomes();
+
+        List<String> expectedChromosomeList = new ArrayList<>();
+        expectedChromosomeList.add("11");
```
pabarcgar: If other tests are not affected, adding some variants for other chromosomes to the test data json file would be nice to have other chromosomes to get a more complex test. If adding more data affects too much other tests, I would trust in the Spring distinct query and wont spend a lot of time here

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
cyenyxe: I suggested using the root myself: the root path lists the segments/chromosomes available (I'm not very happy with the terminology but whatever) and those below that level would refer to "resources" in a segment, like variants.

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryTest.java
tomdcsmith: I've added more variants from other chromosomes now.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
pabarcgar: segments is goof then

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
pabarcgar: It seems that the query including studies can take up to 5 seconds. In that case, and considering that the variants should not be concentrated in just few chromosomes, I think that not having studies filter (like it is now) is better

PR #42, opened 2017-04-19
EVA-599 Use Spring properties files capabilities

tomdcsmith:
- Removed of eva.properties files
- Using Spring's application.properties for holding property values
- @ConfigurationProperties annotation and class for holding application.properties values


eva-lib/src/main/java/uk/ac/ebi/eva/lib/config/EvaProperty.java
```diff
@@ -0,0 +1,130 @@
+package uk.ac.ebi.eva.lib.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+@ConfigurationProperties(prefix = "eva")
+@Component
+public class EvaProperty {
```
cyenyxe: `EvaProperties` would be more aligned with the class name from the standard Java API.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/config/EvaProperty.java
```diff
@@ -0,0 +1,130 @@
...
+
+//    eva.mongo.host=@eva.mongo.host@
+//    eva.mongo.user=@eva.mongo.user@
+//    eva.mongo.passwd=@eva.mongo.passwd@
+//    eva.mongo.auth.db=@eva.mongo.auth.db@
+//    eva.mongo.read-preference=@eva.mongo.read-preference@
+//    eva.mongo.collections.variants=@eva.mongo.collections.variants@
+//    eva.mongo.collections.files=@eva.mongo.collections.files@
+//    eva.version=@eva.version@
```
cyenyxe: Remove commented block, otherwise it should be formatted as class documentation that will have to be updated when the properties file changes.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/config/EvaProperty.java
```diff
@@ -0,0 +1,130 @@
...
+
+    private String version;
+
+    private Mongo mongo;
+
+    public static class Mongo {
+        private String host;
```
cyenyxe: Take advantage of the validation framework to ensure that mandatory fields are not empty among other things. See an example [here](http://blog.codeleak.pl/2014/09/using-configurationproperties-in-spring.html).

eva-lib/src/main/java/uk/ac/ebi/eva/lib/config/EvaProperty.java
cyenyxe: Please look for code usages but I think `eva.version` is no longer necessary.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/config/EvaProperty.java
```diff
@@ -0,0 +1,130 @@
+package uk.ac.ebi.eva.lib.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+@ConfigurationProperties(prefix = "eva")
+@Component
+public class EvaProperty {
+
+//    eva.mongo.host=@eva.mongo.host@
```
cyenyxe: Please review these properties to make them as similar as possible to [those in eva-pipeline](https://github.com/EBIvariation/eva-pipeline/blob/develop/examples/application.properties).

For instance, it should be possible to use Spring Data MongoDB properties instead of our own fields. Further work to inject that automatically could be necessary and I guess could belong in another PR.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/config/EvaProperty.java
```diff
@@ -0,0 +1,130 @@
...
+        private Collections collections;
+
+        public static class Auth {
+            private String db;
+
+            public String getDb() {
+                return db;
+            }
+
+            public void setDb(String db) {
```
cyenyxe: Check if all these setters are still necessary or automatically managed by the framework.

eva-server/src/test/java/uk/ac/ebi/eva/server/configuration/EvaPropertyTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Autowired
+    private EvaProperty evaProperty;
+
+    @Value("${eva.mongo.host}")
+    private String expectedHostProp;
+
+    @Test
+    public void testLoadingOfProperties() {
+        System.out.println("expectedHostProp: " + expectedHostProp);
```
cyenyxe: Not needed, `assertEquals` should accept an extra argument for a message.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/config/EvaProperty.java
```diff
@@ -0,0 +1,130 @@
+package uk.ac.ebi.eva.lib.config;
```
cyenyxe: Please remember the copyright and license headers, they are missing from several new files.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/MongoConfiguration.java
```diff
@@ -29,12 +29,14 @@
...
 
 import uk.ac.ebi.eva.commons.models.converters.data.DBObjectToVariantEntityConverter;
+import uk.ac.ebi.eva.lib.config.EvaProperty;
 
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 
 @Configuration
+@Import(EvaProperty.class)
```
cyenyxe: `DBAdaptorConnector` doesn't need to import this class to inject it, please check when is it really necessary.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/config/EvaProperty.java
cyenyxe: Rename the `config` to `configuration` for consistency with the pipeline project.

eva-server/src/main/java/uk/ac/ebi/eva/server/configuration/EvaPropertyConfiguration.java
```diff
@@ -0,0 +1,9 @@
+package uk.ac.ebi.eva.server.configuration;
+
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ComponentScan(basePackages = {"uk.ac.ebi.eva.lib.config"})
```
cyenyxe: The main `Application` class should scan this package. That would make the component scan a single, centralized step.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/MongoConfiguration.java
tomdcsmith: This is the MongoConfiguration class, did you mistakenly refer to DBAdaptorConnector?
If this import is not there (in MongoConfiguration) then the repository tests fail, complaining there are no beans of type EvaProperties

(i.e.
`Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'uk.ac.ebi.eva.lib.MongoConfiguration': Unsatisfied dependency expressed through field 'evaProperties'; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'uk.ac.ebi.eva.lib.configuration.EvaProperties' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=true)}`
)

eva-lib/src/main/java/uk/ac/ebi/eva/lib/config/EvaProperty.java
tomdcsmith: I've tried removing the setters but the tests fail due to the EvaProperties object not getting instantiated properly- it looks like they are necessary.

eva-server/src/main/java/uk/ac/ebi/eva/server/configuration/EvaPropertyConfiguration.java
tomdcsmith: When I remove this class and add the component scan to the Application class the WSServer tests fail and return 401 errors. I created this class to do this component scan in the same manner as the [MongoEvaLibConfiguration](https://github.com/EBIvariation/eva-ws/blob/develop/eva-server/src/main/java/uk/ac/ebi/eva/server/configuration/MongoEvaLibConfiguration.java) and [JpaRepositoryConfiguration](https://github.com/EBIvariation/eva-ws/blob/develop/eva-server/src/main/java/uk/ac/ebi/eva/server/configuration/JpaRepositoryConfiguration.java) classes.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/config/EvaProperty.java
tomdcsmith: An issue here is that it's not as straightforward as autoconfigured MongoDbFactory, because a basic MongoClient is created [here](https://github.com/EBIvariation/eva-ws/blob/540c5b9cb96a85188604b94d016724632962ad31/eva-lib/src/main/java/uk/ac/ebi/eva/lib/utils/DBAdaptorConnector.java#L87).

install-dependencies.sh
```diff
@@ -4,7 +4,7 @@ git clone -b hotfix/0.4 https://github.com/ebivariation/biodata.git
 git clone -b hotfix/0.5 https://github.com/ebivariation/opencga.git
 git clone -b develop https://github.com/ebivariation/variation-commons.git
 
-cd biodata && mvn install
+cd biodata && git checkout dbfeee0fd551dfb36ac40a2e70d612cf7439f6db && mvn install
```
jmmut: Didn't it work with biodata hotfix/0.4?

tomdcsmith: This is irrelevant now- I've rebased on to develop where this install script isn't needed due to the artifactory.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/configuration/SpringDataMongoDbProperties.java
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.stereotype.Component;
+
+@ConfigurationProperties(ignoreUnknownFields = false, prefix = "spring.data.mongodb")
+@Component
+public class SpringDataMongoDbProperties {
+
+    @NotBlank
+    private String host;
+
+    @NotBlank
```
cyenyxe: This can be blank in any infrastructure that doesn't have user information centralized in a single database, e.g. local deployments.

eva-server/src/main/resources/application.properties
```diff
@@ -19,4 +19,13 @@ springfox.documentation.swagger.v2.path=/webservices/api
...
 
-eva.mongo.collections.files=@eva.mongo.collections.files@
+spring.data.mongodb.host=@eva.mongo.host@
+spring.data.mongodb.authentication-database=@eva.mongo.auth.db@
+spring.data.mongodb.username=@eva.mongo.user@
+spring.data.mongodb.password=@eva.mongo.passwd@
+spring.data.mongodb.read-preference=@eva.mongo.read-preference@
+
+eva.collections.variants=@eva.mongo.collections.variants@
+eva.collections.files=@eva.mongo.collections.files@
```
cyenyxe: These properties are defined differently in the pipeline: those properties are still defined in the pipeline: https://github.com/EBIvariation/eva-pipeline/blob/develop/examples/load-genotyped-vcf.properties#L40-L41

eva-lib/src/main/java/uk/ac/ebi/eva/lib/configuration/EvaProperties.java
```diff
@@ -0,0 +1,70 @@
...
+@Component
+public class EvaProperties {
+
+    private String version;
+
+    @javax.validation.constraints.NotNull
+    private Collections collections;
+
+    public static class Collections {
+        @NotBlank
```
cyenyxe: What about using `javax.validation.constraints.Size` with a minimum size of 1? That would make unnecessary to import a Hibernate package.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/configuration/EvaProperties.java
```diff
@@ -0,0 +1,70 @@
...
+import org.springframework.stereotype.Component;
+import org.springframework.validation.annotation.*;
+
+@ConfigurationProperties(ignoreUnknownFields = false, prefix = "eva")
+@Component
+public class EvaProperties {
+
+    private String version;
+
+    @javax.validation.constraints.NotNull
```
cyenyxe: It should be possible to import this without the full package declaration, there isn't an equally named annotation in `org.hibernate.validator.constraints`: https://docs.jboss.org/hibernate/validator/5.1/api/org/hibernate/validator/constraints/package-summary.html

eva-lib/src/main/java/uk/ac/ebi/eva/lib/configuration/DbCollectionsProperties.java
```diff
@@ -0,0 +1,111 @@
...
+
+@ConfigurationProperties(ignoreUnknownFields = false, prefix = "db.collections")
+@Component
+public class DbCollectionsProperties {
+
+    @Size(min = 1)
+    private String version;
+
+    @NotNull
+    private Files files;
```
cyenyxe: According to the documentation, `In order to validate values of nested properties, you must annotate the associated field as @Valid to trigger its validation.`

eva-lib/src/main/java/uk/ac/ebi/eva/lib/configuration/DbCollectionsProperties.java
```diff
@@ -0,0 +1,111 @@
...
+
+    @NotNull
+    private Variants variants;
```
cyenyxe: Isn't there a way to build this without extra classes, something like a nested namespace? Maybe this just shows how useless the extra level is... Should we drop the `name` suffix from the collection properties in the pipeline instead of adding it to the web services, @jmmut and @jorizci ?

jorizci: Yes, there is no other way of writing that down without the second class. I suggest to reformulate the parameters in both and denormalizate a little bit the configuration parameters.

cyenyxe: Agreed offline to use properties following this pattern:

```
db.collection-names.files
db.collection-names.variants
```

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/ArchiveWSServer.java
```diff
@@ -85,7 +81,7 @@ public QueryResponse countSpecies() {
 
     @RequestMapping(value = "/species/list", method = RequestMethod.GET)
     public QueryResponse getSpecies() {
-        return setQueryResponse(archiveEvaproDbAdaptor.getSpecies(properties.getProperty("eva.version"), true));
+        return setQueryResponse(archiveEvaproDbAdaptor.getSpecies(dbCollectionsProperties.getVersion(), true));
```
cyenyxe: See that the first parameter declared in the [ArchiveEvaproDbAdaptor#getSpecies](https://github.com/EBIvariation/eva-ws/blob/develop/eva-lib/src/main/java/uk/ac/ebi/eva/lib/metadata/ArchiveEvaproDBAdaptor.java#L96) signature is not used. Could you please remove that parameter from the interface and drop the `db.collections.version` property?

eva-lib/src/main/java/uk/ac/ebi/eva/lib/configuration/DbCollectionsProperties.java
cyenyxe: Agreed offline to use properties following this pattern:

```
db.collection-names.files
db.collection-names.variants
```

eva-lib/src/main/java/uk/ac/ebi/eva/lib/metadata/ArchiveEvaproDBAdaptor.java
```diff
@@ -93,7 +89,7 @@ public QueryResult countSpecies() {
     }
 
     @Override
-    public QueryResult getSpecies(String s, boolean b) {
+    public QueryResult getSpecies(boolean b) {
```
cyenyxe: The boolean doesn't seem to be used either?

PR #43, opened 2017-04-19
EVA-509

pabarcgar:
Updated pom.xml to retrieve dependencies from a public repository, and deploy to it. Removed dependencies build script that is not needed by travis anymore


tomdcsmith:
I've tried temporarily removing my local maven repo and when I compile this (after cleaning) it downloads the dependencies without a problem.

PR #44, opened 2017-04-25
EVA-349 Web service for VEP and Ensembl cache used

tomdcsmith:
- Repository for AnnotationMetadata, and tests
- Webservice for AnnotationMetadata, using above repository 
- https://github.com/EBIvariation/variation-commons/pull/29 Needs to be merged first, because this PR relies upon it, which is why the tests are failing for this PR

### DIFF
--- a/eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/AnnotationMetadataRepository.java
+++ b/eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/AnnotationMetadataRepository.java
@@ -21,8 +21,17 @@ import uk.ac.ebi.eva.commons.models.metadata.AnnotationMetadata;
 
 import java.util.List;
 
+/**
+ * Spring MongoRepository for querying collection of AnnotationMetadata documents
+ */
 public interface AnnotationMetadataRepository extends MongoRepository<AnnotationMetadata, String> {
 
+    /**
+     * Query for all AnnotationMetadata from collection, ordered first by cache version descending and then vep version
+     * descending.
+     *
+     * @return List of AnnotationMetadata objects
+     */
     List<AnnotationMetadata> findAllByOrderByCacheVersionDescVepVersionDesc();
 
 }
--- a/eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/AnnotationMetadataRepository.java
+++ b/eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/AnnotationMetadataRepository.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2017 EMBL - European Bioinformatics Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.ac.ebi.eva.lib.repository;
+
+import org.springframework.data.mongodb.repository.MongoRepository;
+
+import uk.ac.ebi.eva.commons.models.metadata.AnnotationMetadata;
+
+import java.util.List;
+
+public interface AnnotationMetadataRepository extends MongoRepository<AnnotationMetadata, String> {
+
+    List<AnnotationMetadata> findAll();
+
+}
--- a/eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/AnnotationMetadataRepository.java
+++ b/eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/AnnotationMetadataRepository.java
@@ -23,6 +23,6 @@ import java.util.List;
 
 public interface AnnotationMetadataRepository extends MongoRepository<AnnotationMetadata, String> {
 
-    List<AnnotationMetadata> findAll();
+    List<AnnotationMetadata> findAllByOrderByCacheVersionDescVepVersionDesc();
 
 }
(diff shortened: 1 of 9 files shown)


tomdcsmith:
Required variation-commons changes deployed to the jfrog artifactory repository https://oss.jfrog.org/artifactory/oss-snapshot-local/uk/ac/ebi/eva/variation-commons/0.2-SNAPSHOT/, allowing build to pass.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/AnnotationMetadataRepository.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public interface AnnotationMetadataRepository extends MongoRepository<AnnotationMetadata, String> {
+
+    /**
+     * Query for all AnnotationMetadata from collection, ordered first by cache version descending and then vep version
+     * descending.
+     *
+     * @return List of AnnotationMetadata objects
+     */
+    List<AnnotationMetadata> findAllByOrderByCacheVersionDescVepVersionDesc();
```
jmmut: Is the cache version more important than the vep version?

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/AnnotationMetadataRepositoryTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Test
+    public void testFindAllByOrderByCacheVersionDescVepVersionDescVepVersionOrder() throws Exception {
+        List<AnnotationMetadata> annotationMetadataList = repository.findAllByOrderByCacheVersionDescVepVersionDesc();
+        AnnotationMetadata prevAnnotationMetadata = annotationMetadataList.get(0);
+        for (AnnotationMetadata curAnnotationMetadata :
+                annotationMetadataList.subList(1, annotationMetadataList.size())) {
+            if (curAnnotationMetadata.getCacheVersion().equals(prevAnnotationMetadata.getCacheVersion())) {
+                assertTrue(Integer.parseInt(curAnnotationMetadata.getCacheVersion())
+                                   <= Integer.parseInt(prevAnnotationMetadata.getCacheVersion()));
```
jmmut: I think you wanted to write here `getVepVersion`

eva-lib/src/test/resources/eva.properties
```diff
@@ -7,6 +7,7 @@ eva.mongo.read-preference=
 
 eva.mongo.collections.variants=vars
 eva.mongo.collections.files=files
+eva.mongo.collections.annotation_metadata=annotation_metadata
```
jmmut: even though they can be different, I would try to homogenize this property names and the ones in the pipeline, @cyenyxe what do you think? I think these are better than the ones in the pipeline ("db.collections.annotation.metadata.name") if we ever use other db backend, and the ".name" feels a bit redundant. Also, I don't know why I split "annotation" and "metadata" with a dot (it's not a different level). Not sure about the underscore case, though.

If we agree to use a common one, as starting point for the discussion, I suggest "eva.mongodb.collections.annotationMetadata"

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/AnnotationMetadataRepository.java
tomdcsmith: Yeah, confirmed by both @garysaunders and @cyenyxe . The updates to the cache are likely to affect the annotation more than the updates to the tool, and updates to the cache are more likely to be changes to the fundamental data.

eva-lib/src/test/resources/eva.properties
tomdcsmith: @cyenyxe and I spoke about this and I have currently changed it in another eva-ws PR (https://github.com/tomdcsmith/eva-ws/blob/479a2d2f60a900b5457740c29324b72f51ec299b/eva-lib/src/test/resources/application.properties) to "db.collections.annotation-metadata.name" for the reason you mention. I think I agree with possibly removing ".name" from the property names.

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/AnnotationMetadataRepositoryTest.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Test
+    public void testFindAllByOrderByCacheVersionDescVepVersionDescCacheVersionOrder() throws Exception {
+        List<AnnotationMetadata> annotationMetadataList = repository.findAllByOrderByCacheVersionDescVepVersionDesc();
+        AnnotationMetadata prevAnnotationMetadata = annotationMetadataList.get(0);
+        for (AnnotationMetadata curAnnotationMetadata :
+                annotationMetadataList.subList(1, annotationMetadataList.size())) {
+            assertTrue(Integer.parseInt(curAnnotationMetadata.getCacheVersion())
+                               <= Integer.parseInt(prevAnnotationMetadata.getCacheVersion()));
```
cyenyxe: If we force this to be a numeric comparison some values will fail:

* 99 < 100 as numbers
* 100 < 99 as strings

So let's just use the string comparison, same as the database will do. Also applies to the method below.

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/AnnotationMetadataRepositoryTest.java
```diff
@@ -0,0 +1,81 @@
...
+        assertEquals(4, annotationMetadataList.size());
+    }
+
+    @Test
+    public void testFindAllByOrderByCacheVersionDescVepVersionDescCacheVersionOrder() throws Exception {
+        List<AnnotationMetadata> annotationMetadataList = repository.findAllByOrderByCacheVersionDescVepVersionDesc();
+        AnnotationMetadata prevAnnotationMetadata = annotationMetadataList.get(0);
+        for (AnnotationMetadata curAnnotationMetadata :
+                annotationMetadataList.subList(1, annotationMetadataList.size())) {
+            assertTrue(0 >= curAnnotationMetadata.getCacheVersion().compareTo(prevAnnotationMetadata.getCacheVersion()));
```
cyenyxe: More idiomatic way below. It makes a bit easier to see the flow because you don't need to remember the value it is compared against while reading the whole `compareTo` call.

```java
curAnnotationMetadata.getCacheVersion().compareTo(prevAnnotationMetadata.getCacheVersion() <= 0
```

PR #45, opened 2017-04-27
EVA-732 File stats in study view returns zero

tomdcsmith:
- DbObjectToVariantGlobalStatsConverter registration in MongoConfiguration
- Test for non-zero stats fields

Requires changes in PR https://github.com/EBIvariation/variation-commons/pull/30


tomdcsmith:
I've deployed the updated variation-commons, so the build is working ok now.

PR #47, opened 2017-05-12
EVA-742 Samples data not displayed

tomdcsmith:
Requires https://github.com/EBIvariation/variation-commons/pull/31/files variation commons PR merging.

To fix issue with samplesData not being populated in variant entities returned from repository (due to sample name information not being present in variant collection.

- Requires [updated DBObjectToSamplesConverter in variation commons](https://github.com/EBIvariation/variation-commons/pull/31), which adds sample data to variant source entries in the variant entity being converted, but uses the samples position/index in the file as the key (rather than the sample name), for the default genotype string it uses just one map entry ("def" -> defaultGtString).
- VariantEntityRepositoryImpl receives variant entities with samplesData in this format.
- VariantEntityRepositoryImpl queries the files collection using VariantSourceEntityRepository, and gets mappings of the sample position/index to the sample name.
- VariantSourceEntityRepository replaces the sample positions/indexes in samplesData with these sample names, it also creates new entries in the samplesData for samples for where the indexes aren't already present (inferred to have the genotype of the default genotype string).

### DIFF
--- a/eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryImpl.java
+++ b/eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryImpl.java
@@ -42,6 +42,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
  * Concrete implementation of the VariantEntityRepository interface (relationship inferred by Spring),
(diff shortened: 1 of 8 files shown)


eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryImpl.java
```diff
@@ -111,12 +149,78 @@ public Long countByRegionsAndComplexFilters(List<Region> regions, List<VariantEn
...
+            }
+            // Put the new sample data into samplesData
+            samplesData.put(sampleName, sampleData);
+        }
+        return variantSourceEntry;
     }
 
     private void addFilterCriteriaToQuery(Query query, List<VariantEntityRepositoryFilter> filters) {
-        if (filters.size() > 0){
+        if ((filters != null) && (filters.size() > 0)){
```
cyenyxe: No need for second level of brackets.

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryTest.java
```diff
@@ -420,12 +421,12 @@ public void testFindByRegionsAndComplexFiltersExcludeAttributes() {
     public void testFindDistinctChromosomesByStudyId() {
         List<String> chromosomeList = variantEntityRepository.findDistinctChromosomes();
```
cyenyxe: Given these are distinct, it would more sense so return a Set instead of a List.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryImpl.java
```diff
@@ -94,6 +102,36 @@ public Long countByRegionsAndComplexFilters(List<Region> regions, List<VariantEn
                                            .distinct("chr");
     }
 
+    private Map<
+                String, Map<
+                    String, Map<
+                        String, String>>> studyIdsToFileIdsToIndexesToNames() {
```
cyenyxe: This private method should be placed after the first usage.

cyenyxe: It would be more efficient if this filtered out non relevant studies.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryImpl.java
```diff
@@ -111,12 +122,108 @@ public Long countByRegionsAndComplexFilters(List<Region> regions, List<VariantEn
...
+
+        variantEntities = updateVariantEntitiesSampleNames(variantEntities, studyIdsToFileIdsToIndexesToNames);
+
+        return variantEntities;
+    }
+
+    private Map<
+            String, Map<
+            String, Map<
+            String, String>>> studyIdsToFileIdsToIndexesToNames() {
```
cyenyxe: The 3-level `Map` could be replaced by a Guava `Table`, which has the same underlying implementation but a much nicer interface: https://github.com/google/guava/wiki/NewCollectionTypesExplained#table

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryTest.java
```diff
@@ -448,6 +449,29 @@ public void testCountByChromosomeAndStartAndTypeAndStudy() {
         assertEquals(new Long(1), count);
     }
 
+    @Test
+    public void testSamplesData() {
```
cyenyxe: `testSamplesDataIsNotEmpty`?

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryTest.java
```diff
@@ -234,7 +235,7 @@ public void testRegionIsFoundWithMafGreaterThan() {
         regions.add(region);
         List<VariantEntityRepositoryFilter> filters = new FilterBuilder().withMaf(">0.125").build();
         List<String> exclude = new ArrayList<>();
-        testFiltersHelperRegion(regions, filters, exclude, 37);
+        testFiltersHelperRegion(regions, filters, exclude, 43);
```
cyenyxe: Have these new numbers being checked against direct queries to the database, in order to make sure the tests themselves are correct?

tomdcsmith: I've now updated the variant test data from the production database, and updated and manually checked all the queries through the command line.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryImpl.java
```diff
@@ -111,12 +124,102 @@ public Long countByRegionsAndComplexFilters(List<Region> regions, List<VariantEn
...
+                continue;
+            }
+
+            String studyId = variantSourceEntity.getStudyId();
+            String fileId = variantSourceEntity.getFileId();
+
+            Map<String, Integer> samplesPosition = variantSourceEntity.getSamplesPosition();
+            Map<String, String> positionSamples = new HashMap<>();
+            for(Map.Entry<String, Integer> entry : samplesPosition.entrySet()) {
+                positionSamples.put(Integer.toString(entry.getValue()), entry.getKey());
```
cyenyxe: Won't `Integer.toString` fail with "def"?

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryImpl.java
```diff
@@ -111,12 +124,102 @@ public Long countByRegionsAndComplexFilters(List<Region> regions, List<VariantEn
...
+                Collectors.toMap(
+                        Map.Entry::getKey,
+                        e -> updateVariantEntitySampleNames(e.getValue(), studyFileIdsToPositionSamples)
+                )
+        );
+        variantEntity.setSourceEntries(variantSourceEntryMap);
+        return variantEntity;
+    }
+
+    private VariantSourceEntry updateVariantEntitySampleNames(VariantSourceEntry variantSourceEntry,
```
cyenyxe: `updateVariantSourceEntrySampleNames`

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryImpl.java
```diff
@@ -111,12 +124,102 @@ public Long countByRegionsAndComplexFilters(List<Region> regions, List<VariantEn
...
+        }
+
+        return studyFileIdsToPositionSamples;
+    }
+
+    private List<VariantEntity> updateVariantEntitiesSampleNames(List<VariantEntity> variantEntities,
+                                                                 Table<String, String, Map<String, String>>
+                                                                         studyFileIdsToPositionSamples) {
+
+        variantEntities = variantEntities.stream().map(
```
cyenyxe: The flow would be more clear if this value was returned directly without assigning it to a input argument.

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryTest.java
```diff
@@ -418,21 +419,21 @@ public void testFindByRegionsAndComplexFiltersExcludeAttributes() {
...
     @Test
     public void testFindDistinctChromosomesByStudyId() {
-        List<String> chromosomeList = variantEntityRepository.findDistinctChromosomes();
+        Set<String> chromosomeSet = variantEntityRepository.findDistinctChromosomes();
 
-        List<String> expectedChromosomeList = new ArrayList<>();
-        expectedChromosomeList.add("11");
-        expectedChromosomeList.add("9");
-        expectedChromosomeList.add("2");
+        Set<String> expectedChromosomeSet = new HashSet<>();
```
cyenyxe: Maybe just `expectedChromosomes` as variable name? It doesn't really matter what the data structure is.

eva-lib/src/test/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryTest.java
```diff
@@ -418,21 +419,21 @@ public void testFindByRegionsAndComplexFiltersExcludeAttributes() {
...
+        expectedChromosomeSet.add("11");
+        expectedChromosomeSet.add("9");
+        expectedChromosomeSet.add("2");
 
-        assertEquals(expectedChromosomeList, chromosomeList);
+        assertEquals(expectedChromosomeSet, new HashSet<>(chromosomeSet));
```
cyenyxe: This cast is no longer necessary.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryImpl.java
tomdcsmith: samplesPosition won't contain "def", samplesPosition is the data from the files collection. If you look at the declaration for samplesPosition `Map<String, Integer> samplesPosition`, the value is guaranteed to be an Integer.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/repository/VariantEntityRepositoryImpl.java
```diff
@@ -111,12 +124,101 @@ public Long countByRegionsAndComplexFilters(List<Region> regions, List<VariantEn
...
+    private List<VariantEntity> updateVariantEntitiesSampleNames(List<VariantEntity> variantEntities,
+                                                                 Table<String, String, Map<String, String>>
+                                                                         studyFileIdsToPositionSamples) {
+
+        return variantEntities.stream().map(
+                variantEntity -> updateVariantSourceEntrySampleNames(variantEntity, studyFileIdsToPositionSamples)
+        ).collect(Collectors.toList());
+    }
+
+    private VariantEntity updateVariantSourceEntrySampleNames(VariantEntity variantEntity,
```
tomdcsmith: For some reason this happened when I used intellij refactoring- it changed both method names.

PR #48, opened 2017-05-15
EVA-757 Migrate variant/exists to using Spring repository

tomdcsmith:
- New VariantEntityRepository queries for finding by coords and study
- Use of these new methods in VariantWSServer for checking variant existence

### DIFF
--- a/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
+++ b/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
@@ -20,11 +20,9 @@
 package uk.ac.ebi.eva.server.ws;
 
 import io.swagger.annotations.Api;
-import org.opencb.biodata.models.feature.Region;
 import org.opencb.datastore.core.QueryResponse;
 import org.opencb.datastore.core.QueryResult;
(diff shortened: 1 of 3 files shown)


eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
```diff
@@ -137,35 +137,58 @@ public QueryResponse checkVariantExists(@PathVariable("variantId") String varian
...
+                                                               regionId[2], alternate);
             }
 
-            QueryResult queryResult = variantMongoDbAdaptor.getAllVariantsByRegion(region, queryOptions);
-            queryResult.setResult(Arrays.asList(queryResult.getNumResults() > 0));
-            queryResult.setResultType(Boolean.class.getCanonicalName());
-            return setQueryResponse(queryResult);
+            numTotalResults = (long) variantEntities.size();
+        } else {
+            response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
```
cyenyxe: We could take this opportunity to make the API more consistent by making this method support searching by ID. It is a bit strange that variants/info works and variants/exists doesn't.

PR #50, opened 2017-06-01
DGVA-64 View list of studies

tomdcsmith:
- New dgva-server package for dgva webservices
- dgva-server using DGVAPRO
- Listing structural studies moved from eva-server to dgva-server, using dgva study browser in DGVAPRO

PR #52, opened 2017-06-05
DGVA-63 Archive summary statistics

tomdcsmith:
- Uses branch from this PR as root: https://github.com/EBIvariation/eva-ws/pull/50, so that PR should be merged first
- Moved study stats webservice for dgva from eva-server to dgva-server
- Moved QueryUtils from EvaWSServer and DgvaWSServer to individual classes
- Helper class for both EVA and DGVA ArchiveWSServer classes for common code


eva-lib/pom.xml
```diff
@@ -75,6 +75,11 @@
             <artifactId>spring-boot-starter-tomcat</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>2.8.4</version>
```
cyenyxe: Version numbers should be removed from children POM files.

eva-lib/src/main/java/uk/ac/ebi/eva/lib/metadata/shared/ArchiveWSServerHelper.java
```diff
@@ -0,0 +1,93 @@
...
+import uk.ac.ebi.eva.lib.metadata.StudyDBAdaptor;
+import uk.ac.ebi.eva.lib.utils.QueryOptions;
+import uk.ac.ebi.eva.lib.utils.QueryResponse;
+import uk.ac.ebi.eva.lib.utils.QueryResult;
+import uk.ac.ebi.eva.lib.utils.QueryUtils;
+
+import java.util.List;
+import java.util.Map;
+
+@Service
```
cyenyxe: Does this really need to be a service? It doesn't seem to be related to Spring, doesn't need to be injected and it just provides some utility methods.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/ArchiveWSServer.java
```diff
@@ -19,138 +19,86 @@
...
     @Autowired
-    private StudyDgvaDBAdaptor studyDgvaDbAdaptor;
-    @Autowired
     private StudyEvaproDBAdaptor studyEvaproDbAdaptor;
+
+    @Autowired
+    private ArchiveWSServerHelper archiveWSServerHelper;
+
+    @Autowired
+    private QueryUtils queryUtils;
```
cyenyxe: After seeing how much repetitive code it adds, I'm not sure that extracting calls to `QueryUtils` from the parent WS class was a good idea :confused: Maybe the 2 commits that changed it that way can be discarded by rebasing?

eva-lib/src/main/java/uk/ac/ebi/eva/lib/metadata/shared/ArchiveWSServerHelper.java
tomdcsmith: It gets injected here https://github.com/tomdcsmith/eva-ws/blob/1693e18546dec145944af16d48cac9f7a923b6ac/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/ArchiveWSServer.java#L56

cyenyxe: But does it need to be? Why not instantiating it in the `ArchiveWSServer` constructor?

tomdcsmith: I think it was @jorizci who suggested using the annotation, I can't remember the specific reason though.

cyenyxe: In this case I don't see any reason to use it. It can incur a performance penalty and it is a completely static dependency, so there is no need for injection.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/ArchiveWSServer.java
tomdcsmith: One of the main reasons they were extracted was because there are now two parent WS classes EvaWSServer in eva-server and DgvaWSServer in dgva-server.

PR #62, opened 2017-08-10
EVA-879 Do not wrap collections in another collection when building response

cyenyxe:
The existing implementation of `setQueryResponse` wrapped anything that was returned in a collection. When the returned value is a collection itself this causes excessive nesting.


tomdcsmith:
I've deployed locally and it's working fine.

PR #66, opened 2017-09-04
EVA-863 Bugfix for genome assembly missing from study pages

cyenyxe:
Assembly accession added to `VariantStudy` model, so it is listed in corresponding endpoints.

### DIFF
--- a/eva-server/src/test/java/uk/ac/ebi/eva/server/ws/ArchiveWSServerTest.java
+++ b/eva-server/src/test/java/uk/ac/ebi/eva/server/ws/ArchiveWSServerTest.java
@@ -109,15 +109,17 @@ public class ArchiveWSServerTest {
         VariantStudy study1 = new VariantStudy("Human Test study 1", "S1", null, "Study 1 description", new int[]{9606},
                                                "Human", "Homo Sapiens", "Germline", "EBI", "DNA", "multi-isolate",
                                                StudyType.CASE_CONTROL, "Exome Sequencing", "ES", "GRCh37",
-                                               "Illumina", new URI("http://www.s1.org"), new String[]{"10"}, 1000, 10);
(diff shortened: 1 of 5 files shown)


eva-server/src/test/java/uk/ac/ebi/eva/server/ws/ArchiveWSServerTest.java
```diff
@@ -109,15 +109,17 @@ public void setup() throws URISyntaxException, IOException {
         VariantStudy study1 = new VariantStudy("Human Test study 1", "S1", null, "Study 1 description", new int[]{9606},
                                                "Human", "Homo Sapiens", "Germline", "EBI", "DNA", "multi-isolate",
                                                StudyType.CASE_CONTROL, "Exome Sequencing", "ES", "GRCh37",
-                                               "Illumina", new URI("http://www.s1.org"), new String[]{"10"}, 1000, 10);
+                                               "GCA_000001405.3", "Illumina", new URI("http://www.s1.org"),
```
pabarcgar: In https://github.com/cyenyxe/eva/blob/c74f35e03984a12d29e6fa460bff499b45f5c3cb/eva-server/src/test/java/uk/ac/ebi/eva/server/ws/ArchiveWSServerTest.java#L221 we were testing if the assembly code was empty, but it seems it was not working, maybe because assembly_chain was null? Is this being tested now?

cyenyxe: Yes, I suspect it is because of `assembly_chain` being null. The thing is that until we drop support for non-accessioned assemblies in EVAPRO there is no way to guarantee that the assembly chain will always be non-null. Once that is done the condition can be removed and the test will fail if this ever happens.

PR #68, opened 2017-09-21
Variant start and end transformed from int to long

pabarcgar:


cyenyxe:
As long as the tests pass once we merge https://github.com/EBIvariation/variation-commons/pull/47, this looks good to me.

PR #70, opened 2017-10-11
EVA-812 Links to submitted and curated FTP in study view page

jmmut:
Add the attribute `browsable` to VariantStudy and EvaStudyBrowser, so that the endpoint meta/studies/all returns studies with a flag telling if it is browsable or not. All browsable studies have at least one file in the table `browsable_file`.

evapro needs to update the materialized view `study_browser` to add 
```
, COALESCE(browsable_table.browsable, false) AS browsable
```
 and 
```
LEFT JOIN ( SELECT browsable_file.project_accession,
            true AS browsable
           FROM browsable_file
          GROUP BY browsable_file.project_accession) browsable_table(project_accession_browsable, browsable) 
  ON browsable_table.project_accession_browsable::text = project.project_accession::text
```

the full command for creating the materialized view is now in eva-lib/src/main/resources/study_browser.sql.


jmmut:
the changes are correct, but github shows wrong commit hashes, probably because of the rebases. the correct commits are 9d2a81e and 5df3d0d.

PR #73, opened 2017-10-20
EVA-933 Added species common name to report

cyenyxe:
Added `commonName` attribute to `ProgressReport` entity.
Made `toVariantWarehouse` and `assemblyFullyMatches` nullable, and other attributes not nullable.


dbsnp-import/src/main/java/uk/ac/ebi/eva/server/models/ProgressReport.java
```diff
@@ -30,19 +31,26 @@
     @Id
     private String databaseName;
 
+    @Column(nullable = false)
     private int taxId;
 
+    @Column(nullable = false)
     private String scientificName;
 
+    @Column(nullable = false)
```
jmmut: should these match the DB? or are these the ones that the DB should have?

cyenyxe: If the DB is created from scratch by the application (test), then it will set the "not null" constraint. If the DB schema already exists (production) it will ensure that null values are never serialized.

PR #80, opened 2018-02-23
EVA-1100 WS running in docker container

pabarcgar:
- Added Dockerfile and other files necessary for the container in the directory 'dockerconf'
- Modified pom.xml to copy the docker files to the same location that eva.war so the container can be built


eva-server/pom.xml
```diff
@@ -98,6 +98,11 @@
                 <directory>src/main/resources</directory>
                 <filtering>true</filtering>
             </resource>
+            <resource>
+                <directory>dockerconf</directory>
```
cyenyxe: Let's use `docker` to follow the same folder structure as in the vcf-validator.

eva-server/dockerconf/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM tomcat:9.0.5-jre8
+ADD eva.war /usr/local/tomcat/webapps/
```
cyenyxe: I thought we agreed to copy the WAR from the target folder? Or shall we leave it as it because it is more flexible?

eva-server/dockerconf/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM tomcat:9.0.5-jre8
```
cyenyxe: The alpine version would help to reduce the image size and make it more suitable for production environments.

eva-server/dockerconf/Dockerfile
pabarcgar: When we build the project, the docker files are copied to the 'target' directory, so when we build the docker image we will be using the WAR from the target folder.

Issue #78, opened 2018-01-30
Improve error codes returned in variant search endpoints


cyenyxe:
Variant search endpoints require 2 annotation version parameters: `annot-vep-version` and `annot-vep-cache-version`. The code currently returned when one of them is not is 500, which represents a server error instead of a user input error. This can be tested by accessing the following endpoints removing one of them: 

https://www.ebi.ac.uk/eva/webservices/rest/v1/segments/5:63323000-63325000/variants?exclude=sourceEntries&species=sscrofa_111&annot-vep-version=89&annot-vep-cache-version=89&page=1&skip=0&limit=10

https://www.ebi.ac.uk/eva/webservices/rest/v1/variants/rs81294174/info?species=sscrofa_111&annot-vep-version=89&annot-vep-cache-version=90

http://wwwdev.ebi.ac.uk/eva/webservices/rest/v1/genes/CD163/variants?species=sscrofa_111&annot-vep-version=89&annot-vep-cache-version=91&page=1&skip=0&limit=10

It would be more appropriate to return a HTTP error code 400, because this is a user input problem.

AdrianBZG:
Hi, I want to work on this issue but it's not clear to me what you mean... Do you mean to check in the AnnotationMetadataWSServer, the list of AnnotationMetadata(s) [2], and check if the vepVersion and cacheVersion of all them are NOT null, if one is NULL, then return error 400, right?

For the version is not found, you mean to check that, if the cacheVersion and vepVersion are not null, they need to have a meaningful value? If so, what is the set of correct values for this fields?

Sorry for the questions, I'm new to EVA repositories.

[1] [https://github.com/EBIvariation/eva-ws/blob/develop/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/AnnotationMetadataWSServer.java](https://github.com/EBIvariation/eva-ws/blob/develop/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/AnnotationMetadataWSServer.java)
[2] [https://github.com/EBIvariation/variation-commons/blob/develop/variation-commons-core/src/main/java/uk/ac/ebi/eva/commons/core/models/AnnotationMetadata.java](https://github.com/EBIvariation/variation-commons/blob/develop/variation-commons-core/src/main/java/uk/ac/ebi/eva/commons/core/models/AnnotationMetadata.java)

cyenyxe:
[...]

PR #84, opened 2018-03-10
Improve error codes returned in variant search endpoints (issue #78)

AdrianBZG:
Improve error codes returned in variant search endpoints #78

### DIFF
--- a/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
+++ b/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
@@ -71,7 +71,7 @@ public class VariantWSServer extends EvaWSServer {
                                         @RequestParam(name = "annot-vep-version", required = false) String annotationVepVersion,
                                         @RequestParam(name = "annot-vep-cache-version", required = false) String annotationVepCacheVersion,
                                         HttpServletResponse response)
-            throws IOException, AnnotationMetadataNotFoundException {
+            throws IOException {
         initializeQuery();
 
         if (annotationVepVersion == null ^ annotationVepCacheVersion == null) {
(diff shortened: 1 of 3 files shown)


AdrianBZG:
Now instead of throwing the AnnotationMetadata exception, which exits the application with internal error (500), it just catches it and returns a 400 error with the message that the AnnotationMetadataException has encapsulated. Hope it's ok, just tell me if I'm getting it wrong, thanks!

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
```diff
@@ -164,27 +169,34 @@ public QueryResponse checkVariantExists(@PathVariable("variantId") String varian
...
+                    variantEntities = queryByCoordinatesAndAlleles(regionId[0], Integer.parseInt(regionId[1]), regionId[2],
                                                                alternate, null, null);
-            }
+                }
 
-        } else {
-            List<VariantRepositoryFilter> filters = new FilterBuilder().withStudies(studies).build();
-            variantEntities = service.findByIdsAndComplexFilters(variantId, filters, null, null,
+            } else {
+                List<VariantRepositoryFilter> filters = new FilterBuilder().withStudies(studies).build();
```
cyenyxe: Please check, but I think only this branch can throw. If so, it would be preferable to surround only that part with a try-catch block, because that way it is clear for the person reading the code that the rest doesn't throw.

AdrianBZG: @cyenyxe Actually, the exception is thrown by 3 branches: line 184 (queryByCoordinatesAndAllelesAndStudyIds), line 187 (queryByCoordinatesAndAlleles), and line 193 (findByIdsAndComplexFilters)... so it's better to leave the try-catch block like it's now? Or maybe use 3 try-catch blocks?

cyenyxe: Okay, I didn't remember those would also throw. Please let it as is then, it will be more clear than having a lot of try-batch blocks.

Issue #77, opened 2018-01-30
Allow searching by a list of variant identifiers


cyenyxe:
The API currently allows to search by a single variant identifier, but it would be useful for a user to input multiple ones in a single request.

This actually requires changes to several repositories:

1. variation-commons: `VariantRepositoryImpl.findByIdsAndComplexFilters` takes a string parameter which should be replaced by a list. In the test class `VariantRepositoryTest.testCountByIdsAndComplexFilters`, there are only a couple tests searching for a single identifier. They should be extended to support a list. The service built on top of the repository class must also be updated to accept a list.
2. eva-ws: The `variant/info` endpoint should automatically split the identifiers by comma.

AdrianBZG:
Hi, the required changes on variation commons to allow to complete this issue are in the following pull request: [https://github.com/EBIvariation/variation-commons/pull/64](https://github.com/EBIvariation/variation-commons/pull/64)

AdrianBZG:
Hi, actually there's a problem, the /info/ endpoint checks if the given variant ID has the symbol ":" in order to query by coordinates and alleles instead of by ids. Modifying the code to handle the new findByIds function returning a List is straigthforward, but I'm not sure about what to do with the region part (splitting by ":").

Let's say the query is composed of 3 variants, and only one of them has the ":", how I merge a feasible result to return using the query by coord+allels for the first one and the new function for the other 2?

Or maybe, should I only take into account checking for the ":" symbol in case that the given list of variants has actually length 1 (only 1 variant...)??

Thanks.

jorizci:
Hi Adrian, can you post a more concrete example of those queries? I'm a little bit removed from this part of the project and my head is mixing some other API we have.

AdrianBZG:
@jorizci Hi:

Let's go briefly of what the current active implementation of {variantId}/info endpoint does:

- You specify a **single** variantId in the route, such as rs199692280 for instance, and get the corresponding variantEntities: [https://github.com/EBIvariation/eva-ws/blob/develop/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java#L120](https://github.com/EBIvariation/eva-ws/blob/develop/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java#L120) **(Type of variant info query 1)**
[...]

PR #86, opened 2018-04-03
Using new func. definition in old endpoint + new endpoint to search b…

AdrianBZG:
…y variant list #77

### DIFF
--- a/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
+++ b/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
@@ -117,11 +117,81 @@ public class VariantWSServer extends EvaWSServer {
                     annotationMetadata = new AnnotationMetadata(annotationVepVersion, annotationVepCacheVersion);
                 }
 
-                variantEntities = service.findByIdsAndComplexFilters(variantId, filters, annotationMetadata, excludeMapped,
+                variantEntities = service.findByIdsAndComplexFilters(Arrays.asList(variantId), filters, annotationMetadata, excludeMapped,
                         Utils.getPageRequest(getQueryOptions()));
 
-                numTotalResults = service.countByIdsAndComplexFilters(variantId, filters);
+                numTotalResults = service.countByIdsAndComplexFilters(Arrays.asList(variantId), filters);
             }
+        } catch (AnnotationMetadataNotFoundException ex) {
+            response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+            return setQueryResponse(ex.getMessage());
+        }
+
+        QueryResult<VariantWithSamplesAndAnnotation> queryResult = buildQueryResult(variantEntities, numTotalResults);
+        return setQueryResponse(queryResult);
+    }
+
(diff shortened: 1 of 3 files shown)


eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
```diff
@@ -131,6 +131,76 @@ public QueryResponse getVariantById(@PathVariable("variantId") String variantId,
         return setQueryResponse(queryResult);
     }
 
+    @RequestMapping(value = "/{variantId}/infoByList", method = RequestMethod.GET)
```
jorizci: Due variantId being a list, I'm not sure if this is the better way of defining the path. I Would use a request parameter insted, @cyenyxe any preferences?

cyenyxe: There is not a lot of consensus about how to do this in the REST community. Given that it is a way to specify multiple resources and it also is how we already do it, I would say we can leave it as is.

jorizci: There is no consensus on this but the use on comma on the parameters is better supported than in the path.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
```diff
@@ -131,6 +131,76 @@ public QueryResponse getVariantById(@PathVariable("variantId") String variantId,
...
+//    @ApiOperation(httpMethod = "GET", value = "Retrieves the information about a variant", response = QueryResponse.class)
```
cyenyxe: Delete this documentation line as it is misleading.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
cyenyxe: Discussed offline and for the sake of consistency with the rest of the existing API we will leave it as is, but I would recommend changing the endpoint the value to `/{variantIds}`

AdrianBZG:
@cyenyxe @jorizci Changes made 😄

jorizci:
For the future: Changes suggested by ... on review is better than "Some changes"

useless_commit_message

AdrianBZG:
@jorizci hahahaha, sorry for that 😆

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServer.java
```diff
@@ -131,6 +131,75 @@ public QueryResponse getVariantById(@PathVariable("variantId") String variantId,
         return setQueryResponse(queryResult);
     }
 
+    @RequestMapping(value = "/{variantIds}/infoByList", method = RequestMethod.GET)
```
cyenyxe: Sorry for the misunderstanding, here I meant that value would be `/{variantIds}` instead of `/{variantIds}/infoByList`

AdrianBZG:
@cyenyxe @jorizci I'm getting error 404 in the test, path not found, but I don't see what it's wrong 😞

PR #90, opened 2018-05-30
EVA-1226 Address dbSNP import progress report usability issues

andresfsilva:
Added fields variantsWithoutEvidenceImported and variantsWithoutEvidenceImportedDate for ProgressReport. This is to be used in dbSNP Import progress.


dbsnp-import/src/test/resources/data.sql
```diff
@@ -1,4 +1,4 @@
-INSERT INTO import_progress (database_name, tax_id, scientific_name, common_name, genbank_assembly_accession, variants_imported, rs_synonyms_imported, in_ensembl, last_dbsnp_build, assembly_fully_matches, to_variant_warehouse, variants_imported_date, rs_synonyms_imported_date) VALUES ('apple_3750', 3750, 'Malus x domestica', 'Apple', 'GCA_000148765.2', 'pending', 'pending', false, 149, false, false, NULL, NULL);
+INSERT INTO import_progress (database_name, tax_id, scientific_name, common_name, genbank_assembly_accession, variants_imported, variants_without_evidence_imported, rs_synonyms_imported, in_ensembl, last_dbsnp_build, assembly_fully_matches, to_variant_warehouse, variants_imported_date, variants_without_evidence_imported_date, rs_synonyms_imported_date) VALUES ('apple_3750', 3750, 'Malus x domestica', 'Apple', 'GCA_000148765.2', 'pending', 'pending', 'pending', false, 149, false, false, NULL, NULL, NULL);
```
pabarcgar: can we add another row to the test data with a non NULL date?

dbsnp-import/src/main/java/uk/ac/ebi/eva/server/models/ProgressReport.java
```diff
@@ -56,12 +56,18 @@
     @Convert(converter = StatusConverter.class)
     private Status variantsImported;
 
+    @Enumerated(EnumType.STRING)
+    @Convert(converter = StatusConverter.class)
+    private Status variantsWithoutEvidenceImported;
```
pabarcgar: I don't like so much the name variantsWithoutEvidenceImported. I cannot think of a btter alternative either... Some suggestions:
- legacyVariantsImported... (but the ones with evidence are also "legacy".)
- unconfirmedVariantsImported 
- unprovedVariatsImported 

This is an internal name anyway, so I'm OK if nobody else can think of a better name.

cyenyxe: We could name the variables the other way around: `variantsWithEvidence` for the ones we already have imported, and just `variants` for them in general. I think it is more clear because `variantsImported` seems to refer to all of them, possibly including those without evidence.

pabarcgar: I like it! That is "darle la vuelta a la tortilla"!

PR #92, opened 2018-06-11
EVA-1226 Address dbSNP import progress report usability issues

andresfsilva:
Added to ProgressReport the fileds: imported_ids, total_ids_dbsnp, imported_syn_ids, total_syn_ids_dbsnp


dbsnp-import/src/main/java/uk/ac/ebi/eva/server/models/ProgressReport.java
```diff
@@ -70,6 +70,14 @@
 
     private Date rsSynonymsImportedDate;
 
+    private Long importedIds;
+
+    private Long totalIdsDbsnp;
+
+    private Long importedSynIds;
+
+    private Long totalSynIdsDbsnp;
```
jmmut: IMO, we shouldn't use abbreviations like Syn

jmmut: Maybe I would drop the last Dbsnp as well, although I can concede it helps understanding that those are the ones in dbsnp, and the imported are in EVA

PR #93, opened 2018-06-20
DGVA-91 DGVA study browser filters

cyenyxe:
DGVa study browser filter by species fixed for comma-separated lists of them. For instance, queries to "chimpanzee" didn't return any results before because this species was always part of studies where multiple primate species were sequenced.


eva-lib/src/main/java/uk/ac/ebi/eva/lib/dgva_utils/DgvaDBUtils.java
```diff
@@ -22,6 +22,7 @@
 import uk.ac.ebi.eva.lib.utils.QueryOptionsConstants;
 
 import static org.springframework.data.jpa.domain.Specifications.where;
+import static uk.ac.ebi.eva.lib.extension.GenericSpecifications.ilike;
 import static uk.ac.ebi.eva.lib.extension.GenericSpecifications.in;
 import static uk.ac.ebi.eva.lib.extension.GenericSpecifications.like;
```
pabarcgar: I think this import is not being used anymore

PR #95, opened 2018-07-11
EVA-1263 Study browser / view pointing to DGVa API

cyenyxe:
Improved display of multi-valued fields from materialized view. Support for assembly names and accessions (previously only names).


eva-lib/src/main/java/uk/ac/ebi/eva/lib/metadata/dgva/ArchiveDgvaDBAdaptor.java
```diff
@@ -48,14 +48,22 @@ public QueryResult countStudiesPerSpecies(QueryOptions queryOptions) {
...
+
         for (Tuple tuple : countGroupBy) {
-            String species = tuple.get(0) != null ? (String) tuple.get(0) : "Others";
+            // Some studies are associated with multiple species, return as comma-separated values
+            String[] species = tuple.get(0) != null ? ((String) tuple.get(0)).split(",") : new String[]{ "Others" };
             long count = (long) tuple.get(1);
-            result.add(new AbstractMap.SimpleEntry<>(species, count));
+            for (String s : species) {
+                if (result.containsKey(s)) {
+                    count += result.get(s);
```
jmmut: why `+=`?

cyenyxe: Because we are dealing with multi-species lists, when they are split the same species could be listed multiple times and we have to accumulate the counts. For instance, the following would count as 2 studies where "human" appears:

* Human
* Chimpanzee, human

PR #96, opened 2018-07-27
EVA-1288 Make Webserivices container runnable in OpenShift

pabarcgar:


tcezard:
EVA-1382 Has been closed without this PR

PR #99, opened 2018-09-27
EVA-1288 Gitlab CI configuration for automatic deployment

pabarcgar:


.gitlab-ci.yml
```diff
@@ -0,0 +1,56 @@
+stages:
```
pabarcgar: In this section, you write the stages for your CI pipeline, in order of execution, top to bottom. There are some predefined stages, but you can write any name here (except reserved words) to define your stages.

.gitlab-ci.yml
```diff
@@ -0,0 +1,56 @@
...
+    - build
+    - deploy
+
+.maven-build:
+    image: maven:3.5.4-jdk-8
```
pabarcgar: Base docker image that will be pulled to execute this job.

.gitlab-ci.yml
```diff
@@ -0,0 +1,56 @@
+stages:
+    - build
+    - deploy
+
+.maven-build:
```
pabarcgar: Pipeline job. The leading dot means the job will be ignored and not executed in the pipeline. This is useful for inheritance, as this can be used as a parent "abstract" job.

.gitlab-ci.yml
```diff
@@ -0,0 +1,56 @@
...
+    image: maven:3.5.4-jdk-8
+    stage: build
```
pabarcgar: stage this job (in this case, the children job of this) will be executed in. All the jobs in the same stage are executed in parallel, unless you define dependencies between them, to make them run sequentially.

.gitlab-ci.yml
```diff
@@ -0,0 +1,56 @@
...
+    variables:
+        ARTIFACT_PATH: "eva-server/target/eva-$ENVIRONMENT_NAME.war"
```
pabarcgar: Convenience variable to make the script commands shorter.

.gitlab-ci.yml
```diff
@@ -0,0 +1,56 @@
...
+maven-build-development:
+    extends: .maven-build
+    variables:
+        MAVEN_PROFILE: "development,evadev,dgvadev"
+        ENVIRONMENT_NAME: "development"
+
+deploy-tomcat-internal:
+    extends: .deploy-tomcat
+    variables:
+        TOMCAT_USER: "$TOMCAT_INTERNAL_USER"
```
pabarcgar: Those variables ($TOMCAT_INTERNAL_USER, ...) are Gitlab CI variables. They can be defined in Gitlab, to store secrets (like credentials).

.gitlab.settings.xml
```diff
@@ -0,0 +1,81 @@
+<settings>
+    <profiles>
+        <profile>
+            <id>internal</id>
+            <properties>
+                <eva.mongo.host>${env.MONGO_HOST_DEVELOPMENT}</eva.mongo.host>
```
pabarcgar: ${env.VARIABLE_NAME} will be substituted with the Gitlab defined variables during the Gitlab CI pipeline execution.

Note the different syntax when using the variables in the .gitlab-ci.yaml file (simple $VARIABLE_NAME in that file)

.gitlab-ci.yml
```diff
@@ -0,0 +1,56 @@
...
+        TOMCAT_PASSWORD: "$TOMCAT_INTERNAL_PASSWORD"
+        TOMCAT_HOST: "$TOMCAT_INTERNAL_HOST"
+        ENVIRONMENT_NAME: "internal"
+
+deploy-tomcat-development:
+    extends: .deploy-tomcat
+    when: manual
```
pabarcgar: This will make the deployment not automatic. If the rest of the jobs in the pipeline are successful, a button will appear in the UI to execute this job manually.

.gitlab-ci.yml
cyenyxe:
```suggestion
# Not executed, parent job definition for Maven builds
.maven-build:
```

.gitlab-ci.yml
```diff
@@ -0,0 +1,56 @@
...
+    image: maven:3.5.4-jdk-8
+    stage: build
+    variables:
+        ARTIFACT_PATH: "eva-server/target/eva-$ENVIRONMENT_NAME.war"
+    script:
+        - "mvn package --settings .gitlab.settings.xml -P $MAVEN_PROFILE"
+        - "cp eva-server/target/eva.war $ARTIFACT_PATH"
+    artifacts:
+        paths:
+            - "$ARTIFACT_PATH"
```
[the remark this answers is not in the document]
cyenyxe: I think we should add the 'expire' attribute now that we have it fresh in our minds, otherwise we will forget about it and end up with lots of artifacts that we don't want to keep. 2 weeks sounds like a reasonable expiration period to me.

.gitlab-ci.yml
```diff
@@ -0,0 +1,93 @@
...
+
+.deploy-tomcat:
```
cyenyxe:
```suggestion
# Not executed, parent job definition for Tomcat deployments
.deploy-tomcat:
```

.gitlab-ci.yml
```diff
@@ -0,0 +1,93 @@
...
+
+.maven-build:
+    image: maven:3.5.4-jdk-8
+    stage: build
+    variables:
+        ARTIFACT_PATH: "eva-server/target/eva-$ENVIRONMENT_NAME.war"
+    script:
+        - "mvn package --settings .gitlab.settings.xml -P $MAVEN_PROFILE"
+        - "cp eva-server/target/eva.war $ARTIFACT_PATH"
+    artifacts:
```
cyenyxe: If we would like to add test reports, we can do it [this way](https://docs.gitlab.com/ee/ci/junit_test_reports.html#maven).

.gitlab-ci.yml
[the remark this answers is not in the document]
cyenyxe:
```suggestion
            - "$ARTIFACT_PATH"
        expire_in: 2 weeks
```

PR #109, opened 2019-03-11
EVA-1465 Bugfix for only one record per database shown in the dbSNP import progress report

cyenyxe:
Added a composite primary key to the `import_progress` table, comprised of the database name and the assembly accession. This allows to have multiple entries per species in those cases where we import multiple assemblies.

Unfortunately this doesn't fully work when combined with HATEOAS support, at least in the version of Spring Data REST that we use. But because the API only needs to list the whole set of species, it has been simplified to do just that.


dbsnp-import/src/main/java/uk/ac/ebi/eva/server/repositories/ProgressReportRepository.java
```diff
@@ -15,23 +15,11 @@
...
-    // Prevents POST, PUT and PATCH
-    @Override
-    @RestResource(exported = false)
-    ProgressReport save(ProgressReport s);
-
-    // Prevents DELETE
-    @Override
-    @RestResource(exported = false)
-    void delete(ProgressReport t);
+public interface ProgressReportRepository extends CrudRepository<ProgressReport, ProgressReportPK> {
```
jmmut: Does this allow post, put, patch and delete?

dbsnp-import/src/main/java/uk/ac/ebi/eva/server/controllers/DbsnpRestController.java
```diff
@@ -0,0 +1,28 @@
+package uk.ac.ebi.eva.server.controllers;
```
cyenyxe: Add copyright and license header.

dbsnp-import/src/main/java/uk/ac/ebi/eva/server/controllers/DbsnpRestController.java
```diff
@@ -0,0 +1,28 @@
...
+
+    private ProgressReportRepository repository;
+
+    public DbsnpRestController(ProgressReportRepository repository) {
+        this.repository = repository;
+    }
+
+    @GetMapping
+    public List<ProgressReport> getReport() {
+        List<ProgressReport> progressReports = Lists.newArrayList(repository.findAll());
```
cyenyxe: Can't we directly return the Iterable?

dbsnp-import/src/main/java/uk/ac/ebi/eva/server/controllers/DbsnpRestController.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import uk.ac.ebi.eva.server.models.ProgressReport;
+import uk.ac.ebi.eva.server.repositories.ProgressReportRepository;
+
+@RestController
+@RequestMapping(value = "/v1/import-status", produces = "application/json")
+public class DbsnpRestController {
```
cyenyxe: Sorry if this is being super-picky, but maybe this could be renamed as `DbsnpProgressReportRestController`? Just to make more clear which dbSNP-related aspect is about, and align it more with the names of the models.

PR #112, opened 2019-05-16
EVA-1525 - Implement rate limiting to prevent server crash

sundarvenkata-EBI:


eva-server/src/main/java/uk/ac/ebi/eva/server/RateLimiterAspect.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(RateLimiterAspect.class);
+
+    private final int RATE_LIMIT_ACQUIRE_TIMEOUT_IN_SECONDS = 10;
+
+    private final ConcurrentHashMap<String, RateLimiter> limiters = new ConcurrentHashMap<>();
+
+    @Before("@annotation(limit)")
+    public void rateLimit(JoinPoint jp, RateLimit limit) throws RateLimitException {
+        RateLimiter limiter = limiters.computeIfAbsent(createKey(jp), createLimiter(limit));
```
jmmut: I had trouble understanding `createKey` and `createLimiter` because they return respectively a value and a function that returns a value:
```suggestion
        String limiterKey = getIpAddress(jp);
        RateLimiter limiter = limiters.computeIfAbsent(limiterKey, getLimiterFactory(limit));
```

eva-server/src/main/java/uk/ac/ebi/eva/server/RateLimiterAspect.java
```diff
@@ -0,0 +1,82 @@
...
+            throw new RateLimitException(String.format("Rate limit exceeded. Please limit rate to %s requests/second.",
+                                                       limit.value()));
+        }
+    }
+
+    private Function<String, RateLimiter> createLimiter(RateLimit limit) {
+        return name -> RateLimiter.create(limit.value());
+    }
+
+    private String createKey(JoinPoint jp) {
```
jmmut: What it does could be used for other things than creating keys
```suggestion
    private String getIpAddress(JoinPoint jp) {
```

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/RegionWSServerIntegrationTest.java
```diff
@@ -139,4 +140,16 @@ public void testVepVersionAndVepCacheVersionFilter() {
...
+    @Test
+    public void testIllegalLimitParameter() {
+        String testRegion = "20:60000-80000";
+        String annotationVepVersion = "78";
+        String annotationVepCacheversion = "78";
+        int limit = 100000;
+        String url = "/v1/segments/" + testRegion +
+                "/variants?species=mmusculus_grcm38&annot-vep-version=" + annotationVepVersion +
+                "&annot-vep-cache-version=" + annotationVepCacheversion + "&limit=" + limit;
+        ResponseEntity response = WSTestHelpers.getRestResponse(url, restTemplate);
```
jmmut:
```suggestion
        ResponseEntity<String> response = restTemplate.getForEntity(url, String.class);
```

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/WSTestHelpers.java
```diff
@@ -61,6 +61,13 @@ public static JSONObject testRestTemplateHelperJsonObject(String url, TestRestTe
         return queryResponse.getResponse().get(0).getResult();
     }
 
+    public static ResponseEntity getRestResponse(String url, TestRestTemplate restTemplate) {
+        return restTemplate.exchange(
+                url, HttpMethod.GET, null,
+                new ParameterizedTypeReference<String>() {
```
jmmut: you don't need a ParameterizedTypeReference for a non-generic type. this whole function is equivalent to restTemplate.getForEntity as explained in my other comment

eva-server/src/test_web_service_rate_limiting.py
```diff
@@ -0,0 +1,79 @@
...
+if __name__ == "__main__":
+    if len(sys.argv) != 2:
+        print("Usage: python3 test_web_service_rate_limiting.py <WEB_SERVICE_HOST_URL> "
+              "(ex: python3 test_web_service_rate_limiting.py http://www.ebi.ac.uk)")
+        sys.exit(1)
+    urlString = "{0}/eva/webservices/rest/v1/segments/1:105000001-105500000/variants?species=mmusculus_grcm38&limit=5"\
+        .format(sys.argv[1])
+    print("To test parallel requests from multiple IP addresses, "
+          "please run this script within 1 minute from other machines ...")
+    time.sleep(60)  # Allow some time for the script to be invoked in multiple machines
```
jmmut: does this really work? the first machine will wait 1 minute, but so will the last one. if the requests are done across 100 seconds, and your delay running the second machine is 30 seconds, the overlap  will be 70 seconds, regardless of how long is this sleep. sleep(0), sleep(60) and sleep(10000) all would yield the same overlap between machines

sundarvenkata-EBI: Added nosleep argument to disable it for other machines

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/RegionWSServerIntegrationTest.java
```diff
@@ -139,4 +140,16 @@ public void testVepVersionAndVepCacheVersionFilter() {
         }
     }
 
+    @Test
+    public void testIllegalLimitParameter() {
+        String testRegion = "20:60000-80000";
+        String annotationVepVersion = "78";
+        String annotationVepCacheversion = "78";
+        int limit = 100000;
+        String url = "/v1/segments/" + testRegion +
```
andresfsilva: I know we have been doing it this way (concatenation the string) but maybe if we use a builder it would be easier to read. Something like:
```
        String url = UriComponentsBuilder.fromUriString("")
                                         .path("/v1/segments/")
                                         .path(testRegion)
                                         .path("/variants")
                                         .queryParam("species", "mmusculus_grcm38")
                                         .queryParam("annot-vep-version", annotationVepVersion)
                                         .queryParam("annot-vep-cache-version", annotationVepCacheversion)
                                         .queryParam("limit", limit)
                                         .build().toString();
```

PR #116, opened 2019-06-12
Variantserverv2

junaidnz97:
Redesigning variant endpoint Apis to follow REST priciples


eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -0,0 +1,227 @@
...
+
+        try {
+            variantEntities = getVariantEntitiesByParams(variantId, annotationVepVersion, annotationVepCacheVersion);
+            numTotalResults = (long) variantEntities.size();
+        } catch (AnnotationMetadataNotFoundException ex) {
+            response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+            return new Resource<>(setQueryResponse(ex.getMessage()));
+        }
+
+        List<VariantWithSamplesAndAnnotation> rootVariantEntities = new ArrayList<>();
```
cyenyxe: Can we build a simpler object than this, given that this endpoint is not going to report samples nor annotation?

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -0,0 +1,227 @@
...
+        });
+
+        QueryResult<VariantWithSamplesAndAnnotation> queryResult = buildQueryResult(rootVariantEntities,
+                numTotalResults);
+
+        Link annotationLink = new Link(linkTo(methodOn(VariantWSServerV2.class).getAnnotations(variantId, species,
+                annotationVepVersion, annotationVepCacheVersion, response)).toUri().toString(), "annotation");
+
+        Link sourceEntriesLink = new Link(linkTo(methodOn(VariantWSServerV2.class).getSourceEntries(variantId, species,
+                annotationVepVersion, annotationVepCacheVersion, response)).toUri().toString(), "sourceEntries");
```
cyenyxe: Shall we rename public references as "sources" instead of using the weirder "sourceEntries"?

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/WSTestHelpers.java
```diff
@@ -61,6 +61,16 @@ public static JSONObject testRestTemplateHelperJsonObject(String url, TestRestTe
         return queryResponse.getResponse().get(0).getResult();
     }
 
+    public static String testRestTemplateHelperForError(String url,
+                                                      TestRestTemplate restTemplate) {
```
cyenyxe: Fix indentation:
```suggestion
                                                        TestRestTemplate restTemplate) {
```

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2Test.java
```diff
@@ -0,0 +1,182 @@
...
+
+    @Autowired
+    private TestRestTemplate restTemplate;
+
+    @MockBean
+    private VariantWithSamplesAndAnnotationsService variantEntityRepository;
+
+    @Before
+    public void setUp() throws Exception {
+
```
cyenyxe:
```suggestion
```

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2IntegrationTest.java
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    @Test
+    public void annotationEndPointTestNonExisting() throws URISyntaxException {
+        String url = "/v2/variants/100:0:C:T/annotations?species=mmusculus_grcm38";
+        ResponseEntity<QueryResponse<QueryResult<Annotation>>> annotations = restTemplate.exchange(
+                url, HttpMethod.GET, null,
+                new ParameterizedTypeReference<QueryResponse<QueryResult<Annotation>>>() {
+                });
+        assertEquals(HttpStatus.OK, annotations.getStatusCode());
```
cyenyxe: In others of our APIs, when an endpoint returns no results, the HTTP code is 404. Applies to this class and `VariantWSServerV2Test`.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
junaidnz97: Okay, I will change it

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2IntegrationTest.java
junaidnz97: I will change it :+1:

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -0,0 +1,227 @@
...
+
+        List<Link> links = new ArrayList<>();
+        links.add(sourceEntriesLink);
+        links.add(annotationLink);
+
+        QueryResponse<QueryResult<VariantSourceEntryWithSampleNames>> sourceEntries = getSourceEntries(variantId,
```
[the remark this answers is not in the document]
junaidnz97: Will fix it in a different PR as discussed offline :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -0,0 +1,227 @@
...
+@Api(tags = {"variants"})
+public class VariantWSServerV2 extends EvaWSServer {
+
+    @Autowired
+    private VariantWithSamplesAndAnnotationsService service;
+
+    @GetMapping(value = "/{variantId}")
+    public Resource<QueryResponse> getCoreInfo(@PathVariable("variantId") String variantId,
+                                               @RequestParam(name = "species") String species,
+                                               @RequestParam(name = "annot-vep-version", required = false)
```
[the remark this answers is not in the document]
junaidnz97: I am sorry, I didn't understand this.
Could you please explain? :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -0,0 +1,227 @@
...
+
+@RestController
+@RequestMapping(value = "/v2/variants", produces = "application/json")
+@Api(tags = {"variants"})
+public class VariantWSServerV2 extends EvaWSServer {
+
+    @Autowired
+    private VariantWithSamplesAndAnnotationsService service;
+
+    @GetMapping(value = "/{variantId}")
```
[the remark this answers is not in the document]
junaidnz97: fixed :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -0,0 +1,227 @@
...
+        links.add(sourceEntriesLink);
+        links.add(annotationLink);
+
+        QueryResponse<QueryResult<VariantSourceEntryWithSampleNames>> sourceEntries = getSourceEntries(variantId,
+                species, annotationVepVersion, annotationVepCacheVersion, response);
+
+        return new Resource<>(setQueryResponse(queryResult), links);
+    }
+
+    private String checkErrorHelper(String variantId, String annotationVepVersion, String annotationVepCacheVersion,
```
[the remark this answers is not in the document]
junaidnz97: fixed :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
junaidnz97: Fixed :+1:

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -0,0 +1,227 @@
...
+@RequestMapping(value = "/v2/variants", produces = "application/json")
+@Api(tags = {"variants"})
+public class VariantWSServerV2 extends EvaWSServer {
+
+    @Autowired
+    private VariantWithSamplesAndAnnotationsService service;
+
+    @GetMapping(value = "/{variantId}")
+    public Resource<QueryResponse> getCoreInfo(@PathVariable("variantId") String variantId,
+                                               @RequestParam(name = "species") String species,
```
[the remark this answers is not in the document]
junaidnz97: Fixed :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
junaidnz97: I have created a new class named VariantCoreInfo. Would that be okay?

jmmut: Check if uk.ac.ebi.eva.commons.core.models.pipeline.Variant works for this before creating another bean

junaidnz97: I tried using this now, but the `Type` and `length` field doesn't get retrieved along with the response as there is no datamember of the same in `uk.ac.ebi.eva.commons.core.models.pipeline.Variant` class

eva-server/pom.xml
```diff
@@ -79,6 +79,11 @@
             <artifactId>guava</artifactId>
             <version>27.0-jre</version>
         </dependency>
+        <dependency>
```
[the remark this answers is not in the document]
junaidnz97: I tried making this change, but the response type changed adding some new field and retrieving the null fields as well along with the response probably. I guess I can try to get it right in the next PR since we would be changing the response type by removing QueryResponse and QueryResult there anyway.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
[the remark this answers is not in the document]
cyenyxe: What I mean is when you search for a particular variant `1:1000:A:C`, the version of the annotation is not important for retrieving the core variant information.
The only reason why we could need the related parameters (`annot-vep-version` and `annot-vep-cache-version`) would be to construct the HATEOAS links to the same annotation as requested, instead of displaying them as generic ones. But I think it is preferrable to show them as generic, and including a separate endpoint that returns which versions are available.

Is this of any help? :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -54,67 +54,74 @@
...
-        QueryResponse<QueryResult<VariantSourceEntryWithSampleNames>> sourceEntries = getSourceEntries(variantId,
-                species, annotationVepVersion, annotationVepCacheVersion, response);
-
         return new Resource<>(setQueryResponse(queryResult), links);
     }
 
-    private String checkErrorHelper(String variantId, String annotationVepVersion, String annotationVepCacheVersion,
-                                    String species) {
-        if (!variantId.contains(":")) {
+    private String checkParameters(String variantCoreString, String annotationVepVersion,
```
cyenyxe: Part of this method throws `IllegalArgumentException` and another part returns the error message. Please make all branches throw an exception.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantCoreInfo.java
```diff
@@ -0,0 +1,39 @@
...
+
+    private String alternate;
+
+    private Set<String> ids;
+
+    private VariantType type;
+
+    private int length;
+
+    private Map<String, Set<String>> hgvs;
```
cyenyxe: Our HGVS algorithm is pretty primitive, so I would prefer to exclude that field from the response.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
cyenyxe: The problem comes from [this section of the code](https://github.com/EBIvariation/eva-ws/blob/master/eva-server/src/main/java/uk/ac/ebi/eva/server/configuration/JacksonConfiguration.java#L50-L54), which only searches for attributes, not getter methods. We had to do this for the REST API v1 because the model classes are complex, and they have a lot of getters that provide different views to the same information. But the `Variant` class should not give any problems to be retrieved via getter methods.

So we have 2 choices:

1. Create 2 different `ObjectMapper`s and configure them based on the URL (`/v1` for fields and `/v2` for getters). I don't know how we can do this but I guess it is possible.
2. Create a new Maven module that only contains the API v2, making the `ObjectMapper` configuration straightforward.

junaidnz97: I will check this out and will get back to you :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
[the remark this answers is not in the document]
junaidnz97: Yeah :+1: 
Fixed :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
junaidnz97: Fixed by using `objectMapper.addMixIn(Variant.class, VariantMixin.class)` :+1:

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
junaidnz97: Fixed :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/AnnotationMetadataWSServerV2.java
```diff
@@ -0,0 +1,59 @@
...
+    public List<AnnotationMetadata> getAnnotationMetadata(@RequestParam(name = "species") String species,
+                                                          @RequestParam(name = "assembly") String assembly)
+            throws IllegalArgumentException {
+        if (species.isEmpty()) {
+            throw new IllegalArgumentException("Please specify a species");
+        }
+        if(assembly.isEmpty()) {
+            throw new IllegalArgumentException("Please specify an assembly");
+        }
+        MultiMongoDbFactory.setDatabaseNameForCurrentThread(DBAdaptorConnector.getDBName(species+"_"+assembly));
```
cyenyxe:
```suggestion
        MultiMongoDbFactory.setDatabaseNameForCurrentThread(DBAdaptorConnector.getDBName(species + "_" + assembly));
```

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -55,20 +56,17 @@
...
+                                @RequestParam(name = "assembly") String assembly,
+                                HttpServletResponse response) throws IllegalArgumentException{
         initializeQuery();
-        String errorMessage = checkParameters(variantCoreString, annotationVepVersion, annotationVepCacheVersion,
-                species, assembly);
-        if (errorMessage != null) {
+        try {
+            checkParameters(variantCoreString, null, null, species, assembly);
+        } catch (IllegalArgumentException e) {
+            System.out.println(e.getMessage());
```
cyenyxe: Although logging is nice in production, it should be implemented with a logger class, instead of just printing into the standard output.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -0,0 +1,240 @@
...
+
+        MultiMongoDbFactory.setDatabaseNameForCurrentThread(DBAdaptorConnector.getDBName(species + "_" + assembly));
+
+        List<VariantWithSamplesAndAnnotation> variantEntities;
+        Long numTotalResults;
+
+        try {
+            variantEntities = getVariantEntitiesByParams(variantCoreString, null, null);
+            numTotalResults = (long) variantEntities.size();
+            if (variantEntities.size() == 0) {
```
cyenyxe:
```suggestion
            if (numTotalResults == 0L) {
```

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -0,0 +1,240 @@
...
+            annotations.add(variantEntity.getAnnotation());
+        });
+
+        QueryResult<Annotation> queryResult = buildQueryResult(annotations, annotations.size());
+
+        return setQueryResponse(queryResult);
+    }
+
+    @GetMapping(value = "/{variantCoreString}/source-entries")
+    public QueryResponse getsources(@PathVariable("variantCoreString") String variantCoreString,
```
cyenyxe: Rename this method as `getSources`. Also, if the HATEOAS link is called `sources`, it would be reasonable to make the URL `/{variantCoreString}/sources` too.

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/AnnotationMetadataWSServerV2Test.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Test
+    public void testFetAnnotationMetadata() {
+        String url = "/v2/annotation-metadata?species=mmusculus&assembly=grcm38";
+        ResponseEntity<List<AnnotationMetadata>> response = restTemplate.exchange(
+                url, HttpMethod.GET, null,
+                new ParameterizedTypeReference<List<AnnotationMetadata>>() {
+                });
+        assertEquals(HttpStatus.OK, response.getStatusCode());
+        System.out.println(response);
```
cyenyxe:
```suggestion
```

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/AnnotationMetadataWSServerV2Test.java
```diff
@@ -0,0 +1,75 @@
...
+    @Before
+    public void setUp() throws Exception {
+        annotationMetadataList = Arrays.asList(new AnnotationMetadata("75", "75"),
+                new AnnotationMetadata("74", "74"));
+
+        given(service.findAllByOrderByCacheVersionDescVepVersionDesc()).willReturn(annotationMetadataList);
+    }
+
+    @Test
+    public void testFetAnnotationMetadata() {
```
cyenyxe:
```suggestion
    public void testGetAnnotationMetadata() {
```

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/AnnotationMetadataWSServerV2.java
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.web.bind.annotation.RestController;
+import uk.ac.ebi.eva.commons.core.models.AnnotationMetadata;
+import uk.ac.ebi.eva.commons.mongodb.services.AnnotationMetadataService;
+import uk.ac.ebi.eva.lib.eva_utils.DBAdaptorConnector;
+import uk.ac.ebi.eva.lib.eva_utils.MultiMongoDbFactory;
+
+import java.util.List;
+
+@RestController
+@RequestMapping(value = "/v2/annotation-metadata", produces = "application/json")
```
cyenyxe: Should we make this URL `/v2/annotation-versions` to make its purpose more obvious? Or even open the door to show info about a species or assembly with `/v2/species/{species}/assembly/{assembly}/annotations`? The latter looks cool but we may prefer that for a hypothetical v3, when everything is more stable.

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/AnnotationMetadataWSServerV2Test.java
```diff
@@ -0,0 +1,75 @@
...
+        String url = "/v2/annotation-metadata?species=mmusculus&assembly=grcm38";
+        ResponseEntity<List<AnnotationMetadata>> response = restTemplate.exchange(
+                url, HttpMethod.GET, null,
+                new ParameterizedTypeReference<List<AnnotationMetadata>>() {
+                });
+        assertEquals(HttpStatus.OK, response.getStatusCode());
+        System.out.println(response);
+        List<AnnotationMetadata> annotationMetadataList= response.getBody();
```
cyenyxe:
```suggestion
        List<AnnotationMetadata> annotationMetadataList = response.getBody();
```

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2IntegrationTest.java
```diff
@@ -0,0 +1,175 @@
...
+    @Autowired
+    private TestRestTemplate restTemplate;
+    @Autowired
+    private VariantWithSamplesAndAnnotationsService service;
+
+    @Before
+    public void setUp() throws Exception { }
+
+    @Test
+    public void rootTestGetVariantsByVariantId() throws URISyntaxException {
```
cyenyxe:
```suggestion
    public void rootTestGetVariantsByVariantCoreString() throws URISyntaxException {
```

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2IntegrationTest.java
```diff
@@ -0,0 +1,175 @@
...
+        List<VariantWithSamplesAndAnnotation> variantWithSamplesAndAnnotations = WSTestHelpers.testRestTemplateHelper(
+                url,restTemplate);
+        assertTrue(variantWithSamplesAndAnnotations.size() > 0);
+        assertTrue(variantWithSamplesAndAnnotations.get(0).getSourceEntries().size() == 0);
+        assertNull(variantWithSamplesAndAnnotations.get(0).getAnnotation());
+        assertTrue(variantWithSamplesAndAnnotations.get(0).getIds().size() > 0);
+    }
+
+    @Test
+    public void rootTestGetVariantsByNonExistingVariantId() throws URISyntaxException {
```
cyenyxe:
```suggestion
    public void rootTestGetVariantsByNonExistingVariantCoreString() throws URISyntaxException {
```

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2IntegrationTest.java
```diff
@@ -0,0 +1,175 @@
...
+    @Before
+    public void setUp() throws Exception { }
+
+    @Test
+    public void rootTestGetVariantsByVariantId() throws URISyntaxException {
+        String url = "/v2/variants/20:60100:A:T?species=mmusculus&assembly=grcm38";
+        List<VariantWithSamplesAndAnnotation> variantWithSamplesAndAnnotations = WSTestHelpers.testRestTemplateHelper(
+                url,restTemplate);
+        assertTrue(variantWithSamplesAndAnnotations.size() > 0);
+        assertTrue(variantWithSamplesAndAnnotations.get(0).getSourceEntries().size() == 0);
```
cyenyxe: This looks more natural to me, either that or `assertEquals(0, ...)`.
```suggestion
        assertTrue(variantWithSamplesAndAnnotations.get(0).getSourceEntries().isEmpty());
```

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2IntegrationTest.java
```diff
@@ -0,0 +1,175 @@
...
+        ResponseEntity<QueryResponse<QueryResult<VariantWithSamplesAndAnnotation>>> response = restTemplate.exchange(
+                url, HttpMethod.GET, null,
+                new ParameterizedTypeReference<QueryResponse<QueryResult<VariantWithSamplesAndAnnotation>>>() {
+                });
+        assertEquals(HttpStatus.NOT_FOUND, response.getStatusCode());
+        assertEquals(0, response.getBody().getResponse().size());
+    }
+
+    @Test
+    public void rootTestForError() throws URISyntaxException {
```
cyenyxe: Make the name a bit more specific, and add another assert for a case when the assembly is not provided, or none of them are.
```suggestion
    public void rootTestParameterErrors() throws URISyntaxException {
```

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2IntegrationTest.java
```diff
@@ -0,0 +1,175 @@
...
+        ResponseEntity<QueryResponse<QueryResult<Annotation>>> annotations = restTemplate.exchange(
+                url, HttpMethod.GET, null,
+                new ParameterizedTypeReference<QueryResponse<QueryResult<Annotation>>>() {
+                });
+        assertEquals(HttpStatus.OK, annotations.getStatusCode());
+        assertFalse(annotations.getBody().getResponse().get(0).getResult().get(0).getChromosome().isEmpty());
+    }
+
+    @Test
+    public void annotationEndPointTestNonExisting() throws URISyntaxException {
```
cyenyxe: Different situations of non-existence should be checked: one where the variant does not exist, and another where the variant exists but it is not annotated.

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2IntegrationTest.java
```diff
@@ -0,0 +1,175 @@
...
+        ResponseEntity<QueryResponse<QueryResult<Annotation>>> annotations = restTemplate.exchange(
+                url, HttpMethod.GET, null,
+                new ParameterizedTypeReference<QueryResponse<QueryResult<Annotation>>>() {
+                });
+        assertEquals(HttpStatus.NOT_FOUND, annotations.getStatusCode());
+        assertTrue(annotations.getBody().getResponse().get(0).getResult().size() == 0);
+    }
+
+    @Test
+    public void annotationEndpointTestForError() throws URISyntaxException {
```
cyenyxe: Add another assert for the case when the cache version is provided but the VEP version is not.

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2IntegrationTest.java
```diff
@@ -0,0 +1,175 @@
...
+                exchange(url, HttpMethod.GET, null,
+                        new ParameterizedTypeReference<QueryResponse<QueryResult
+                                <VariantSourceEntryWithSampleNames>>>() {
+                        });
+        assertEquals(HttpStatus.OK, annotations.getStatusCode());
+        assertFalse(annotations.getBody().getResponse().get(0).getResult().get(0).getFileId().isEmpty());
+    }
+
+    @Test
+    public void sourceEntriesEndPointTestNonExisting() throws URISyntaxException {
```
cyenyxe: Same as in the method for testing annotations, it should check the case where the variant doesn't exist, or when it exists and the statistics don't.

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2Test.java
```diff
@@ -0,0 +1,181 @@
...
+    private static final String MAIN_ID = "rs1";
+
+    private static final VariantWithSamplesAndAnnotation VARIANT = new VariantWithSamplesAndAnnotation("1", 1000, 1005,
+            "A", "T", MAIN_ID);
+
+    @Autowired
+    private TestRestTemplate restTemplate;
+
+    @MockBean
+    private VariantWithSamplesAndAnnotationsService variantEntityRepository;
```
cyenyxe: Service or repository?

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2Test.java
```diff
@@ -0,0 +1,181 @@
...
+        assertTrue(variantWithSamplesAndAnnotations.get(0).getIds().size() > 0);
+    }
+
+    private List<VariantWithSamplesAndAnnotation> variantWsHelper(String testVariantId) {
+        String url = "/v2/variants/" + testVariantId + "?species=mmusculus&assembly=grcm38";
+        return WSTestHelpers.testRestTemplateHelper(url, restTemplate);
+    }
+
+    @Test
+    public void rootTestGetVariantsByNonExistingVariantId() throws URISyntaxException {
```
cyenyxe:
```suggestion
    public void rootTestGetVariantsByNonExistingVariantCoreString() throws URISyntaxException {
```

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2Test.java
```diff
@@ -0,0 +1,181 @@
...
+        VARIANT.addSourceEntry(new VariantSourceEntryWithSampleNames("fid", "sid", null, null, null, null, null));
+        List<VariantWithSamplesAndAnnotation> variantEntities = Collections.singletonList(VARIANT);
+
+        given(variantEntityRepository
+                .findByChromosomeAndStartAndReferenceAndAlternate(eq(CHROMOSOME), anyInt(), any(), any(), any()))
+                .willReturn(variantEntities);
+    }
+
+    @Test
+    public void rootTestGetVariantsByVariantId() throws URISyntaxException {
```
cyenyxe:
```suggestion
    public void rootTestGetVariantsByVariantCoreString() throws URISyntaxException {
```

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/AnnotationMetadataWSServerV2.java
jmmut: in the future annotation-metadata could contain other info than the versions, although I can't think of anything to put there, so I'm fine with /v2/annotation-versions.

I'm not sure I like `/v2/species/{species}/assembl*ies*/{assembly}/annotations`, as it kind of suggests that species are containers of assemblies, which doesn't sound elegant to me.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
junaidnz97: Sorry, I did this for debugging and missed to remove it.
I have removed this :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
junaidnz97: Fixed :+1:

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/AnnotationMetadataWSServerV2.java
junaidnz97: I have changed this to `/v2/annotation-versions` for now :)

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2Test.java
junaidnz97: Service.
Fixed :)

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2IntegrationTest.java
junaidnz97: Added :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -0,0 +1,240 @@
...
+            throw new IllegalArgumentException("Please specify an assembly");
+        }
+    }
+
+    private List<VariantWithSamplesAndAnnotation> getVariantEntitiesByParams(String variantCoreString,
+                                                                             String annotationVepVersion,
+                                                                             String annotationVepCacheVersion) throws
+            AnnotationMetadataNotFoundException {
+        String[] regionId = variantCoreString.split(":");
+        String alternate = (regionId.length > 3) ? regionId[3] : null;
```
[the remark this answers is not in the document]
junaidnz97: I have added this.
Would this be okay?

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -127,25 +130,29 @@ private void checkParameters(String variantCoreString, String annotationVepVersi
...
+            String variantCoreString,
+                                                                                 String annotationVepVersion,
+                                                                                 String annotationVepCacheVersion)
+            throws AnnotationMetadataNotFoundException, IllegalArgumentException {
         String[] regionId = variantCoreString.split(":");
-        String alternate = (regionId.length > 3) ? regionId[3] : null;
-
+        Optional<String> alternate = Optional.ofNullable(regionId[3]);
+        if (!alternate.isPresent()) {
+            return Optional.ofNullable(null);
```
jmmut: no, I meant that the method *requires* the 4 fields. if the split by ":" doesn't give 4 fields, then throw an exception asking for them.

Try to imagine as a user what would happen with the code you made. you would ask for chr:start:ref, and what would you get? a 404, not found. is that what really happened? does that response explain how the user can make their request work? :)

junaidnz97: Oh, So should I take in `chromosome, start, reference and alternate` as individual request parameters?

junaidnz97: I have added the check in the checkParameters functiom :)

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2IntegrationTest.java
junaidnz97: I am a little bit confused regarding how to add this test since `/test-data/variants.json` file and both the documents in that have `annotation` and `source entries` fields present.
Is it okay to add extra data to this file to satisfy the requirement?

cyenyxe: Sure, it is totally valid according to our schema to have a variant without annotation. Variants should always have source entries though, the only part that could be missing are the statistics.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -59,47 +60,47 @@
...
-            numTotalResults = (long) variantEntities.size();
-            if (variantEntities.size() == 0) {
-                response.setStatus(HttpServletResponse.SC_NOT_FOUND);
-                return new Resource<>(setQueryResponse(variantEntities));
-            }
+            variantEntity = getVariantByCoordinatesAndAnnotationVersion(variantCoreString, null, null);
         } catch (AnnotationMetadataNotFoundException ex) {
             response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
             return new Resource<>(setQueryResponse(ex.getMessage()));
+        } catch (IllegalArgumentException ex) {
```
cyenyxe: If multiple exception types are treated the same way, Java allows to capture them all as described [here](http://tutorials.jenkov.com/java-exception-handling/catching-multiple-exceptions.html).

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -59,47 +60,47 @@
...
-                    variantEntity.getEnd(), variantEntity.getReference(),variantEntity.getAlternate());
-            variant.setIds(variantEntity.getIds());
-            variantList.add(variant);
-        });
+        if (!variantEntity.isPresent()) {
+            response.setStatus(HttpServletResponse.SC_NOT_FOUND);
+            return new Resource<>(setQueryResponse(variantList));
+        }
+
+        Variant variant = new Variant(variantEntity.get().getChromosome(), variantEntity.get().getStart(),
```
cyenyxe: In order to avoid many redundant calls, extract the result from `variantEntity.get()` to a local variable, which could be called something like `retrievedVariant`.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -109,9 +110,20 @@ public Resource getCoreInfo(@PathVariable("variantCoreString") String variantCor
     }
 
     private void checkParameters(String variantCoreString, String annotationVepVersion,
-                                 String annotationVepCacheVersion, String species, String assembly)  {
+                                 String annotationVepCacheVersion, String species, String assembly) {
         if (!variantCoreString.contains(":")) {
             throw new IllegalArgumentException("Please describe a variant as 'sequence:location:reference:alternate'");
+        } else {
```
cyenyxe: This doesn't need to be nested under an `else` branch, as throwing an exception will interrupt the method execution. In fact, I think both checks are a bit redundant, because the split + length check effectively covers the contains check too.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -0,0 +1,257 @@
...
+        return new Resource<>(setQueryResponse(queryResult), links);
+    }
+
+    private void checkParameters(String variantCoreString, String annotationVepVersion,
+                                 String annotationVepCacheVersion, String species, String assembly) {
+        if (!variantCoreString.contains(":")) {
+            throw new IllegalArgumentException("Please describe a variant as 'sequence:location:reference:alternate'");
+        } else {
+            String[] regionId = variantCoreString.split(":");
+            if(regionId.length!=4) {
```
cyenyxe:
```suggestion
            if (regionId.length != 4) {
```

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -109,9 +110,20 @@ public Resource getCoreInfo(@PathVariable("variantCoreString") String variantCor
...
 
     private void checkParameters(String variantCoreString, String annotationVepVersion,
-                                 String annotationVepCacheVersion, String species, String assembly)  {
+                                 String annotationVepCacheVersion, String species, String assembly) {
         if (!variantCoreString.contains(":")) {
             throw new IllegalArgumentException("Please describe a variant as 'sequence:location:reference:alternate'");
+        } else {
+            String[] regionId = variantCoreString.split(":");
+            if(regionId.length!=4) {
+                throw new IllegalArgumentException("VariantCoreString requires 4 fields, " + regionId.length +
```
cyenyxe: I think a more clear message would be again `Please describe a variant as 'sequence:location:reference:alternate'`

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -0,0 +1,257 @@
...
+                        "fields were given in the input");
+            }
+            try {
+                Integer.parseInt(regionId[1]);
+            } catch (Exception e) {
+                throw new IllegalArgumentException("Please specify a valid Integer for `start` in VariantCoreString");
```
cyenyxe:
```suggestion
                throw new IllegalArgumentException("Please specify a valid integer start position");
```

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -0,0 +1,257 @@
...
+            throw new IllegalArgumentException("Please describe a variant as 'sequence:location:reference:alternate'");
+        } else {
+            String[] regionId = variantCoreString.split(":");
+            if(regionId.length!=4) {
+                throw new IllegalArgumentException("VariantCoreString requires 4 fields, " + regionId.length +
+                        "fields were given in the input");
+            }
+            try {
+                Integer.parseInt(regionId[1]);
+            } catch (Exception e) {
```
cyenyxe: Be as specific as possible when declaring which exceptions are caught. This guarantees that if an unexpected exception is thrown, it will have the appropriate visibility.
```suggestion
            } catch (NumberFormatException e) {
```

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -0,0 +1,257 @@
...
+        if (!variantCoreString.contains(":")) {
+            throw new IllegalArgumentException("Please describe a variant as 'sequence:location:reference:alternate'");
+        } else {
+            String[] regionId = variantCoreString.split(":");
+            if(regionId.length!=4) {
+                throw new IllegalArgumentException("VariantCoreString requires 4 fields, " + regionId.length +
+                        "fields were given in the input");
+            }
+            try {
+                Integer.parseInt(regionId[1]);
```
cyenyxe: We use long numbers to represent a position, and it should be a positive number.
```suggestion
                long position = Long.parseLong(regionId[1]);
                if (position < 0) {
                    throw new IllegalArgumentException("Please provide a positive start position");
                }
```

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -0,0 +1,257 @@
...
+        }
+
+        List<VariantWithSamplesAndAnnotation> variantWithSamplesAndAnnotationList = service.
+                findByChromosomeAndStartAndReferenceAndAlternate(regionId[0], Integer.parseInt(regionId[1]),
+                        regionId[2], regionId[3], annotationMetadata);
+
+        if (variantWithSamplesAndAnnotationList.size() == 1) {
+            return Optional.of(variantWithSamplesAndAnnotationList.get(0));
+        } else if (variantWithSamplesAndAnnotationList.size() > 1) {
+            throw new IllegalArgumentException("More than one variant retireved. Please enter a unique VariantId");
```
cyenyxe:
```suggestion
            throw new IllegalArgumentException("More than one variant has been found.");
```

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -169,31 +184,33 @@ public QueryResponse getAnnotations(@PathVariable("variantCoreString") String va
...
+            variantEntity = getVariantByCoordinatesAndAnnotationVersion(variantCoreString, annotationVepVersion,
                     annotationVepCacheVersion);
-            if (variantEntities.size() == 0) {
-                response.setStatus(HttpServletResponse.SC_NOT_FOUND);
-                return setQueryResponse(buildQueryResult(variantEntities, 0));
-            }
         } catch (AnnotationMetadataNotFoundException ex) {
             response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
             return setQueryResponse(ex.getMessage());
+        } catch (IllegalArgumentException ex) {
```
cyenyxe: Again, multiple exception types that are handled identically can be grouped in a single `catch` block.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -212,26 +229,26 @@ public QueryResponse getsources(@PathVariable("variantCoreString") String varian
...
```
cyenyxe: Group exception types handled identically in a single `catch` block.

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2IntegrationTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+    @Before
+    public void setUp() throws Exception { }
+
+    @Test
+    public void rootTestGetVariantsByVariantCoreString() throws URISyntaxException {
+        String url = "/v2/variants/20:60100:A:T?species=mmusculus&assembly=grcm38";
+        List<VariantWithSamplesAndAnnotation> variantWithSamplesAndAnnotations = WSTestHelpers.testRestTemplateHelper(
+                url,restTemplate);
+        assertTrue(variantWithSamplesAndAnnotations.size() == 1);
```
cyenyxe: The common idiom would be:
```suggestion
        assertEquals(1, variantWithSamplesAndAnnotations.size());
```

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2Test.java
```diff
@@ -0,0 +1,180 @@
...
+        given(service.findByChromosomeAndStartAndReferenceAndAlternate(eq(CHROMOSOME), anyInt(), any(), any(), any()))
+                .willReturn(variantEntities);
+    }
+
+    @Test
+    public void rootTestGetVariantsByVariantCoreString() throws URISyntaxException {
+        String url = "/v2/variants/" + CHROMOSOME +":71822:C:G?species=mmusculus&assembly=grcm38";
+        List<VariantWithSamplesAndAnnotation> variantWithSamplesAndAnnotations = WSTestHelpers
+                .testRestTemplateHelper(url,restTemplate);
+        assertTrue(variantWithSamplesAndAnnotations.size() == 1);
```
cyenyxe:
```suggestion
        assertTrue(1, variantWithSamplesAndAnnotations.size());
```

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2IntegrationTest.java
junaidnz97: Okay, I will add it :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
junaidnz97: Fixed :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
junaidnz97: Fixed :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
junaidnz97: Fixed :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
junaidnz97: Fixed :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
junaidnz97: Fixed :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
junaidnz97: Fixed :+1:

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2IntegrationTest.java
junaidnz97: I have added this :)

eva-server/pom.xml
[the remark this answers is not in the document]
cyenyxe: Do you know in which version of Spring HATEOAS was this fixed? I can see that, if we upgrade to Spring Boot 1.5.7 like other projects, the starter will bring HATEOAS 0.23.0.RELEASE, see [here](http://central.maven.org/maven2/org/springframework/boot/spring-boot-dependencies/1.5.7.RELEASE/spring-boot-dependencies-1.5.7.RELEASE.pom).

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -0,0 +1,247 @@
...
+                                 String annotationVepCacheVersion, String species, String assembly) {
+        String[] regionId = variantCoreString.split(":");
+        if (regionId.length != 4) {
+            throw new IllegalArgumentException("Please describe a variant as 'sequence:location:reference:" +
+                    "alternate'");
+        }
+        try {
+            long position = Long.parseLong(regionId[1]);
+            if (position < 0) {
+                throw new IllegalArgumentException("Please provide a positive start position");
```
cyenyxe: 0 is a valid position used to describe a very particular situation.
```suggestion
                throw new IllegalArgumentException("Please provide a start position equals or greater than zero");
```

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2IntegrationTest.java
```diff
@@ -125,14 +126,31 @@ public void annotationEndPointTestExisting() throws URISyntaxException {
     }
 
     @Test
-    public void annotationEndPointTestNonExisting() throws URISyntaxException {
+    public void annotationEndPointTestNonExistingWithNonExistingVariant() throws URISyntaxException {
```
cyenyxe: I don't understand the name of this method. If you are trying to test one case where only the variant does not exist, and another where neither the variant nor annotation exist, you can create two methods and name each one more specifically.

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2IntegrationTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+    @Test
+    public void sourceEntriesEndPointTestNonExistingWithNonExistingVariant() throws URISyntaxException {
+        String url = "/v2/variants/100:0:C:T/sources?species=mmusculus&assembly=grcm38";
+        ResponseEntity<QueryResponse<QueryResult<VariantSourceEntryWithSampleNames>>> sources = restTemplate.
+                exchange(url, HttpMethod.GET, null, new ParameterizedTypeReference<QueryResponse<
+                        QueryResult<VariantSourceEntryWithSampleNames>>>() {
+                });
+        assertEquals(HttpStatus.NOT_FOUND, sources.getStatusCode());
+        assertTrue(sources.getBody().getResponse().get(0).getResult().size() == 0);
```
cyenyxe:
```suggestion
        assertTrue(sources.getBody().getResponse().get(0).getResult().isEmpty());
```

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2IntegrationTest.java
junaidnz97: Fixed :)

eva-server/pom.xml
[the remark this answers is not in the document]
junaidnz97: I will check if it is working with HATEOAS 0.23.0.RELEASE now :)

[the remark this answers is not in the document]
junaidnz97: I checked this for `0.23.0.RELEASE` and `0.24.0.RELEASE` versions and it is not working.
But it is working for 0.25.0.RELEASE

PR #122, opened 2019-07-17
Improve variant ws server v2

junaidnz97:
The purpose of this PR is to remove the usage of  `QueryResponse` in VariantWSServerV2 and associated tests.

### DIFF
--- a/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
+++ b/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
@@ -21,6 +21,8 @@ package uk.ac.ebi.eva.server.ws;
 
 import io.swagger.annotations.Api;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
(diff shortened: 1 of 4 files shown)


eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -51,22 +53,20 @@
...
 @RequestMapping(value = "/v2/variants", produces = "application/json")
 @Api(tags = {"variants"})
-public class VariantWSServerV2 extends EvaWSServer {
+public class VariantWSServerV2  {
 
     @Autowired
     private VariantWithSamplesAndAnnotationsService service;
 
     @GetMapping(value = "/{variantCoreString}")
-    public Resource getCoreInfo(@PathVariable("variantCoreString") String variantCoreString,
```
jmmut: I thought we were going to use hateoas in all the v2 endpoints, weren't we?

junaidnz97: Yes, I am returning 
` return new ResponseEntity(new Resource<>(variant, links),HttpStatus.OK);`
So the output format would still be the same.
Use of ResponseEntity would help to return the cases where error happens easily, I guess

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/VariantWSServerV2.java
```diff
@@ -185,40 +178,32 @@ public QueryResponse getAnnotations(@PathVariable("variantCoreString") String va
...
         }
 
-        List<Annotation> annotations = new ArrayList<>();
         if (!variantEntity.isPresent() || variantEntity.get().getAnnotation() == null) {
-            response.setStatus(HttpServletResponse.SC_NOT_FOUND);
-            return setQueryResponse(buildQueryResult(annotations, 0));
+            return new ResponseEntity(null, HttpStatus.NOT_FOUND);
         }
+        Link coreVariantLink = new Link(linkTo(methodOn(VariantWSServerV2.class).getCoreInfo(variantCoreString,
+                species, assembly, response)).toUri().toString(), "CoreVariant");
```
jmmut: try to keep the case consistent with every other endpoint, let's use camelCase
```suggestion
                species, assembly, response)).toUri().toString(), "coreVariant");
```

PR #124, opened 2019-07-22
bug fix and optimization for beaconquery

junaidnz97:
The purpose of this PR is to add checks so that querying of imprecise queries is only allowed when all of startMin,startMax,endMin and endMax are present.

Also, in the case where `includeDatasetResponses` is `None`, an optimization is done by not calling `findByRegionAndOtherBeaconFilters`

### DIFF
--- a/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/ga4gh/beaconv2/BeaconServiceV2.java
+++ b/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/ga4gh/beaconv2/BeaconServiceV2.java
@@ -100,7 +100,7 @@ public class BeaconServiceV2 {
 
         try {
             checkParameters(chromosome, referenceBases, start, end, alternateBases, variantType, assemblyId,
-                    includeDatasetResponses);
+                    includeDatasetResponses, startMin, startMax, endMin, endMax);
         } catch (IllegalArgumentException e) {
             return buildBeaconAlleleResponse(null, null, null, e.getMessage());
         }
@@ -139,7 +139,8 @@ public class BeaconServiceV2 {
 
     private void checkParameters(String chromosome, String referenceBases, Long start, Long end,
                                  String alternateBases, String variantType, String assemblyId,
-                                 String includeDatasetResponses) throws IllegalArgumentException {
+                                 String includeDatasetResponses, Long startMin, Long startMax, Long endMin,
+                                 Long endMax) throws IllegalArgumentException {
         if (chromosome == null || chromosome.length() == 0) {
             String errorMessage = "Please provide a valid reference name from ";
             throw new IllegalArgumentException(errorMessage + String.join(", ",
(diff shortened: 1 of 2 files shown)


eva-server/src/main/java/uk/ac/ebi/eva/server/ws/ga4gh/beaconv2/BeaconServiceV2.java
```diff
@@ -161,11 +162,18 @@ private void checkParameters(String chromosome, String referenceBases, Long star
...
+                throw new IllegalArgumentException("Please provied either start or both startMin and startMax");
+            }
+        } else if (start < 0) {
             throw new IllegalArgumentException("Please provide a positive start number");
         }
-
-        if (end != null && end < 0) {
+        if (end == null) {
+            if (endMin == null || endMax == null) {
+                throw new IllegalArgumentException("Please provied either end or both endMin and endMax");
```
jmmut: > the use of "start" without an "end" parameter requires the use of "referenceBases"

but it's allowed not to provide any end* parameter

junaidnz97: I have removed this check for end parameter :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/ga4gh/beaconv2/BeaconServiceV2.java
```diff
@@ -161,10 +166,13 @@ private void checkParameters(String chromosome, String referenceBases, Long star
             throw new IllegalArgumentException("Please provide reference bases");
         }
 
-        if (start != null && start < 0) {
+        if (start == null) {
```
andresfsilva: If a `start` is provided we need to check there is either an `end` or a `reference`

If a `start` is not provided (imprecise positions) then we need to check all fields `startMin` + `startMax` + `endMin` + `endMax` are provided

There are some missing checks here

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/ga4gh/beaconv2/BeaconServiceV2.java
```diff
@@ -139,7 +143,8 @@ public BeaconAlleleResponse find(String chromosome, Long start, Long startMin, L
 
     private void checkParameters(String chromosome, String referenceBases, Long start, Long end,
                                  String alternateBases, String variantType, String assemblyId,
-                                 String includeDatasetResponses) throws IllegalArgumentException {
+                                 String includeDatasetResponses, Long startMin, Long startMax, Long endMin,
```
andresfsilva: Fields `endMin` and `endMax` are not used in this method and they need to be checked when receiving imprecise queries

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/ga4gh/beaconv2/BeaconServiceV2.java
junaidnz97: Actually `referenceBases` is a parameter that is always required and check for that is done just above this, so I don;t think that it is needed to be checked again.
I will add the check for endMin and endMax here :)

junaidnz97: Added :)

PR #125, opened 2019-07-25
Add pagination to region server v2

junaidnz97:
The purpose of this PR is to add pagination feature to RegionWSServerV2 endpoint

### DIFF
--- a/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServerV2.java
+++ b/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServerV2.java
@@ -171,14 +171,12 @@ public class RegionWSServerV2 {
 
     private PageMetadata buildPageMetadata(Integer pageSize, Integer pageNumber, Integer totalNumberOfResults)
             throws IllegalArgumentException {
-
         Long totalPages = pageSize == 0L ? 0L : (long) Math.ceil((double) totalNumberOfResults / (double) pageSize);
 
         if (pageNumber < 0 || pageNumber >= totalPages) {
             throw new IllegalArgumentException("For the given page size, there are " + totalPages + " page(s), so the" +
                     " correct page range is from 0 to " + String.valueOf(totalPages - 1) + " (both included).");
         }
-
         return new PagedResources.PageMetadata(pageSize, pageNumber, totalNumberOfResults, totalPages);
     }
 
@@ -202,7 +200,6 @@ public class RegionWSServerV2 {
 
             resourcesList.add(new Resource<>(variant, Arrays.asList(sourcesLink, annotationsLink)));
         });
-
         return resourcesList;
     }
 
@@ -241,11 +238,10 @@ public class RegionWSServerV2 {
                                       String siftScore, String annotationVepVersion, String annotationVepCacheVersion,
                                       int pageNumber, int pageSize, HttpServletResponse response,
                                       HttpServletRequest request, String linkName) {
-        return new Link(linkTo(methodOn(RegionWSServerV2.class).getVariantsByRegion(regionId, species,
-                assembly, studies, consequenceType, maf, polyphenScore, siftScore, annotationVepVersion,
+        return new Link(linkTo(methodOn(RegionWSServerV2.class).getVariantsByRegion(regionId, species, assembly,
+                studies, consequenceType, maf, polyphenScore, siftScore, annotationVepVersion,
                 annotationVepCacheVersion, pageNumber, pageSize, response, request))
                 .toUriComponentsBuilder()
                 .toUriString(), linkName);
     }
-
 }
--- a/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServerV2.java
+++ b/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServerV2.java
@@ -69,6 +69,10 @@ public class RegionWSServerV2 {
     @Autowired
     private VariantWithSamplesAndAnnotationsService service;
 
+    private static final String SIZE= "size";
+
+    private static final String PAGE_NUMBER = "page";
+
     public RegionWSServerV2() {
     }
 
@@ -151,8 +155,8 @@ public class RegionWSServerV2 {
                     annotationVepCacheVersion, new PageRequest(pageable.getPageNumber() - 1, pageable.getPageSize()),
                     response, request))
                     .toUriComponentsBuilder()
-                    .queryParam("size", pageable.getPageSize())
-                    .queryParam("page", pageable.getPageNumber() - 1)
+                    .queryParam(SIZE, pageable.getPageSize())
+                    .queryParam(PAGE_NUMBER, pageable.getPageNumber() - 1)
                     .toUriString(), "prev"));
 
             pagedResources.add(new Link(linkTo(methodOn(RegionWSServerV2.class).getVariantsByRegion(regionId, species,
@@ -160,8 +164,8 @@ public class RegionWSServerV2 {
                     annotationVepCacheVersion, new PageRequest(0, pageable.getPageSize()),
                     response, request))
                     .toUriComponentsBuilder()
-                    .queryParam("size", pageable.getPageSize())
-                    .queryParam("page", 0)
+                    .queryParam(SIZE, pageable.getPageSize())
+                    .queryParam(PAGE_NUMBER, 0)
                     .toUriString(), "first"));
         }
         if (pageable.getPageNumber() + 1 != pageMetadata.getTotalPages()) {
@@ -170,8 +174,8 @@ public class RegionWSServerV2 {
                     annotationVepCacheVersion, new PageRequest(pageable.getPageNumber() + 1, pageable.getPageSize()),
                     response, request))
                     .toUriComponentsBuilder()
-                    .queryParam("size", pageable.getPageSize())
-                    .queryParam("page", pageable.getPageNumber() + 1)
+                    .queryParam(SIZE, pageable.getPageSize())
+                    .queryParam(PAGE_NUMBER, pageable.getPageNumber() + 1)
                     .toUriString(), "next"));
 
             pagedResources.add(new Link(linkTo(methodOn(RegionWSServerV2.class).getVariantsByRegion(regionId, species,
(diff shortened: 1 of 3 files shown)


eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServerV2.java
```diff
@@ -137,7 +142,49 @@ public ResponseEntity getVariantsByRegion(@PathVariable("regionId") String regio
...
+                    .queryParam("page", pageable.getPageNumber() + 1)
+                    .toUriString(), "next"));
+
+            pagedResources.add(new Link(linkTo(methodOn(RegionWSServerV2.class).getVariantsByRegion(regionId, species,
+                    assembly, studies, consequenceType, maf, polyphenScore, siftScore, annotationVepVersion,
+                    annotationVepCacheVersion, new PageRequest(
+                            (int)pageMetadata.getTotalPages()-1, pageable.getPageSize()),
+                    response, request))
+                    .toUriComponentsBuilder()
+                    .queryParam("size", pageable.getPageSize())
```
jmmut: extract at the very minimum those strings that appear several times. The strings that appear in this class once but are used from other classes should arguably be extracted too.

junaidnz97: I have changed `size` and `page` to constant Strings :+1:

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServerV2.java
```diff
@@ -137,7 +152,49 @@ public ResponseEntity getVariantsByRegion(@PathVariable("regionId") String regio
 
             resourcesList.add(new Resource<>(variant, Arrays.asList(sourcesLink, annotationsLink)));
         });
-        return new ResponseEntity(new Resources<>(resourcesList), HttpStatus.OK);
+
+        PagedResources pagedResources = new PagedResources<>(resourcesList, pageMetadata);
+
+        if (pageable.getPageNumber() > 0) {
+            pagedResources.add(new Link(linkTo(methodOn(RegionWSServerV2.class).getVariantsByRegion(regionId, species,
```
andresfsilva: When there is code that is repetitive with just some differences try to extract it to a method with the needed parameters, that way the code will be more readable. In this you could use `createPaginationLink` and use it for creating the first, last, next and previous links

junaidnz97: Fixed :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServerV2.java
```diff
@@ -92,12 +95,19 @@ public ResponseEntity getVariantsByRegion(@PathVariable("regionId") String regio
                                                       annotationVepVersion,
                                               @RequestParam(name = "annot-vep-cache-version", required = false) String
                                                       annotationVepCacheVersion,
-                                              Pageable pageable,
+                                              @RequestParam(value = "page", required = false) Integer page,
+                                              @RequestParam(value = "size", required = false) Integer size,
```
jmmut: looking at the other parameters, size might get confused with region size, so I suggest renaming to pageSize

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/RegionWSServerV2IntegrationTest.java
```diff
@@ -136,4 +140,37 @@ public void testGetVariantsByNonExistingRegion() throws URISyntaxException {
...
+        assertEquals(10, size.intValue());
+        assertEquals(1, totalPages.intValue());
+    }
+
+    @Test
+    public void testInvalidPageRanges() {
+        String url = "/v2/regions/20:60000-62000/variants?species=mmusculus&assembly=grcm38?&page=1000&size=1";
+        ResponseEntity<String> response = restTemplate.getForEntity(url, String.class);
+        assertEquals(HttpStatus.REQUESTED_RANGE_NOT_SATISFIABLE, response.getStatusCode());
+        assertEquals("The correct page range is from 0 to 0 for the given page size", response.getBody());
```
jmmut: this is perfectly correct, but looks suspicious. can you change it so that it shows like this?
```suggestion
        assertEquals("For the given page size, there are 1 page(s), so the correct page range is from 0 to 0 (both included).", response.getBody());
```

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServerV2.java
```diff
@@ -87,11 +91,19 @@ public ResponseEntity getVariantsByRegion(@PathVariable("regionId") String regio
...
                                               @RequestParam(name = "annot-vep-cache-version", required = false) String
                                                       annotationVepCacheVersion,
+                                              @RequestParam(value = "page", required = false) Integer page,
+                                              @RequestParam(value = "size", required = false) Integer size,
                                               HttpServletResponse response,
                                               @ApiIgnore HttpServletRequest request)
             throws IllegalArgumentException {
         checkParameters(annotationVepVersion, annotationVepCacheVersion, species);
 
+        if (page == null) {
```
andresfsilva: You can set the deafult value in the `RequestParam` annotation [link](https://docs.spring.io/spring/docs/4.3.14.RELEASE/javadoc-api/org/springframework/web/bind/annotation/RequestParam.html)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServerV2.java
```diff
@@ -102,21 +114,31 @@ public ResponseEntity getVariantsByRegion(@PathVariable("regionId") String regio
...
 
         List<VariantWithSamplesAndAnnotation> variantEntities;
         List<Resource> resourcesList = new ArrayList<>();
 
-        if (pageSize == 0) {
+        if (totalNumberOfResults == 0) {
             response.setStatus(HttpServletResponse.SC_NOT_FOUND);
-            return new ResponseEntity(new Resources<>(resourcesList), HttpStatus.NOT_FOUND);
+            return new ResponseEntity(new PagedResources<>(resourcesList, new PageMetadata(size, page < 0 ? 0 : page,
+                    totalNumberOfResults)), HttpStatus.NOT_FOUND);
```
jmmut: same as with the genes endpoint, an empty list is more a 204 "no content" than a 404 "not found"

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServerV2.java
junaidnz97: Fixed :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServerV2.java
junaidnz97: Fixed :+1:

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServerV2.java
junaidnz97: Fixed :)

andresfsilva: Delete the constants `DEFAULT_PAGE_SIZE` and `DEFAULT_PAGE` if they will not be used

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServerV2.java
andresfsilva: probably renaming `page` to `pageNumber` could also be a good idea to be consistent

junaidnz97: I have renamed `page` to `pageNumber` :+1:

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServerV2.java
junaidnz97: Deleted :)

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServerV2.java
```diff
@@ -87,6 +87,10 @@ public ResponseEntity getVariantsByRegion(@PathVariable("regionId") String regio
                                                       annotationVepVersion,
                                               @RequestParam(name = "annot-vep-cache-version", required = false) String
                                                       annotationVepCacheVersion,
+                                              @RequestParam(value = "page", required = false, defaultValue = "0")
+                                                      Integer pageNumber,
+                                              @RequestParam(value = "size", required = false, defaultValue = "20")
```
jmmut: I was talking about renaming the external parameter `value = "size"` :joy:

junaidnz97: I have changed the external parameter name as well now :sweat_smile:

eva-server/src/test/java/uk/ac/ebi/eva/server/ws/RegionWSServerV2Test.java
```diff
@@ -123,6 +124,10 @@ private void testGetVariantsByRegionHelper(String testRegion, int expectedVarian
         ResponseEntity<String> response = restTemplate.getForEntity(url, String.class);
         assertEquals(status, response.getStatusCode());
 
+        if (status == HttpStatus.NO_CONTENT) {
+            assertNull(response.getBody());
+            return null;
+        }
```
cyenyxe: For increased clarity, because this completely breaks the method flow.
```suggestion
        }

```

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServerV2.java
```diff
@@ -137,7 +151,29 @@ public ResponseEntity getVariantsByRegion(@PathVariable("regionId") String regio
 
             resourcesList.add(new Resource<>(variant, Arrays.asList(sourcesLink, annotationsLink)));
         });
-        return new ResponseEntity(new Resources<>(resourcesList), HttpStatus.OK);
+
+        PagedResources pagedResources = new PagedResources<>(resourcesList, pageMetadata);
```
cyenyxe: I would extract the whole construction of the `PagedResources` object to an auxiliary `buildPage` method. In fact, I think this whole method could be split in 3 clear parts, and each of them becoming one:

1. Build page metadata
2. Get resources
3. Build page

junaidnz97: Fixed :)

PR #129, opened 2019-08-22
BugFixes

junaidnz97:
The purpose of this PR is to perform bug fixes for v2 APIs

### DIFF
--- a/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/GeneWSServerV2.java
+++ b/eva-server/src/main/java/uk/ac/ebi/eva/server/ws/GeneWSServerV2.java
@@ -104,7 +104,7 @@ public class GeneWSServerV2 {
             HttpServletResponse response,
             @ApiIgnore HttpServletRequest request)
             throws IllegalArgumentException {
-        checkParameters(species, assembly, bufferValue);
+        checkParameters(geneIds, species, assembly, bufferValue);
         MultiMongoDbFactory.setDatabaseNameForCurrentThread(DBAdaptorConnector.getDBName(species + "_" + assembly));
         List<FeatureCoordinates> featureCoordinates = service.findAllByGeneIdsOrGeneNames(geneIds, geneIds);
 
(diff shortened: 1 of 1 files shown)


eva-server/src/main/java/uk/ac/ebi/eva/server/ws/GeneWSServerV2.java
```diff
@@ -137,7 +137,14 @@ public ResponseEntity getVariantsByGene(
                 response, request), HttpStatus.OK);
     }
 
-    private void checkParameters(String species, String assembly, Integer bufferValue) throws IllegalArgumentException {
+    private void checkParameters(List<String> geneIds, String species, String assembly, Integer bufferValue)
+            throws IllegalArgumentException {
+        long emptyGeneIdCount = geneIds.stream().map(String::isEmpty).filter(value -> value == true).count();
```
jmmut: what this is checking is if any element is an empty string. I didn't think about this but it would be nice to search only non-empty strings, but my comment was that the endpoint allows to search by an empty *list*, which this method will allow right now.

eva-server/src/main/java/uk/ac/ebi/eva/server/ws/GeneWSServerV2.java
```diff
@@ -137,7 +137,17 @@ public ResponseEntity getVariantsByGene(
...
 
-    private void checkParameters(String species, String assembly, Integer bufferValue) throws IllegalArgumentException {
+    private void checkParameters(List<String> geneIds, String species, String assembly, Integer bufferValue)
+            throws IllegalArgumentException {
+        List<String> filteredGeneIds = geneIds.stream().filter(geneId -> geneId.isEmpty() == false).
+                collect(Collectors.toList());
+        geneIds.clear();
+        geneIds.addAll(filteredGeneIds);
+
+        if (geneIds.size() == 0) {
```
andresfsilva:
```suggestion
        if (geneIds.isEmpty()) {
```

PR #133, opened 2020-02-21
EVA-1104 - Take out mappings for 2KB upstream and downstream variants per VEP

sundarvenkata-EBI:
After fix, most severe consequence type appears correctly: https://wwwdev.ebi.ac.uk/eva/?Variant-Browser&species=ecaballus_20&selectFilter=region&region=1%3A260000-270000


sundarvenkata-EBI:
Updated to use LinkedHashMap

PR #135, opened 2020-05-05
EVA-1914 Use the maven settings from github private repository in gitlab

andresfsilva:
- Use maven settings from the github private repo 
- Changes to be consistent with the gitlab pipeline done in `eva-accession`
- Set default-domain in application.properties


eva-server/src/main/resources/application.properties
```diff
@@ -33,3 +33,5 @@ db.collection-names.annotations=@eva.mongo.collections.annotations@
 
 management.endpoints.web.exposure.include=info,health
 management.info.git.mode=full
+
+spring.jmx.default-domain=eva.ebi.ac.uk.@timestamp@
```
sundarvenkata-EBI, line 37: I am wondering why not keep this consistent with the approach [here](https://github.com/EBIvariation/eva-accession/blob/b3836f9e6516ab68660b958f13a7db437660ace2/eva-accession-ws/src/main/resources/application.properties#L26) and use pipe symbol?

.gitlab-ci.yml
```diff
@@ -1,113 +1,123 @@
...
-    script:
-        - "mvn package --settings .gitlab.settings.xml -P $MAVEN_PROFILE"
-        - "cp eva-server/target/eva-*.war $ARTIFACT_PATH"
-    artifacts:
-        paths:
-            - "$ARTIFACT_PATH"
-        expire_in: 2 weeks
+variables:
+    WS_ARTIFACT_PATH: eva-server/target/eva-$ENVIRONMENT_NAME.war
+    MAVEN_SETTINGS: maven-settings.xml
```
sundarvenkata-EBI, line 8:
```suggestion
    MAVEN_SETTINGS_XML: maven-settings.xml
```

eva-server/src/main/resources/application.properties
andresfsilva: In the `eva-accession` project we [changed the default delimiter](https://github.com/EBIvariation/eva-accession/blob/master/pom.xml#L144) because we were having some issues but it was not the case for the `eva-ws` project

.gitlab-ci.yml
```diff
@@ -1,113 +1,123 @@
...
     when: manual
     variables:
-        TOMCAT_USER: "$TOMCAT_PRODUCTION_FALLBACK_USER"
-        TOMCAT_PASSWORD: "$TOMCAT_PRODUCTION_FALLBACK_PASSWORD"
-        TOMCAT_HOST: "$TOMCAT_PRODUCTION_FALLBACK_HOST"
-        ENVIRONMENT_NAME: "production"
+        ENVIRONMENT_NAME: production
+        TOMCAT_USER: $TOMCAT_PRODUCTION_FALLBACK_USER
+        TOMCAT_PASSWORD: $TOMCAT_PRODUCTION_FALLBACK_PASSWORD
+        TOMCAT_HOST: $TOMCAT_PRODUCTION_FALLBACK_HOST
```
sundarvenkata-EBI, line 119: Why aren't the double quotes needed any more?

eva-server/src/main/resources/application.properties
jmmut: the pipe symbol was configured for eva-accession because you can't use `@` as replacement if a property is a URL that contains `@`, which was needed for connecting to mongo through URL. 

https://github.com/EBIvariation/eva-accession/blob/master/pom.xml#L145 

https://github.com/EBIvariation/eva-accession/blob/master/eva-accession-ws/src/main/resources/application.properties#L13

Here in eva-ws we don't use the connection to mongo through URL so the default is used.

.gitlab-ci.yml
andresfsilva: I actually prefer `MAVEN_SETTINGS` plus it will be consistent with what we have in the `eva-accession` pipeline

.gitlab-ci.yml
andresfsilva: Quotes were used before but weren't really needed. I think it looks better without them so I removed them

PR #137, opened 2020-05-20
EVA-1902 - Spring Boot 2 upgrade

sundarvenkata-EBI:
- Replace Fongo with real MongoDB tests because Fongo is not supported in Spring Boot 2.1 - see [here](https://github.com/fakemongo/fongo/issues/357)
  - Update Travis and Gitlab scripts to include Mongo installations
  - Update POMs to support "|" delimiters so that EVA mongo host for tests can be parametrized for both Travis and Gitlab
- javax.validation API dependency need no longer be explicitly pulled in since Spring Boot 2 contains it
- Upgrade jackson-databind version to 2.9.7 to work with Spring Boot 2
- JPA repository query methods should be marked with "Query" annotation


eva-server/src/test/java/uk/ac/ebi/eva/server/test/rule/FixSpringMongoDbRule.java
```diff
@@ -0,0 +1,24 @@
+package uk.ac.ebi.eva.server.test.rule;
+
+import com.lordofthejars.nosqlunit.mongodb.MongoDbConfiguration;
+import com.lordofthejars.nosqlunit.mongodb.SpringMongoDbRule;
+
+/**
+ * Temporary fix until nosql unit rc-6 or final is released
+ */
+public class FixSpringMongoDbRule extends SpringMongoDbRule {
```
jmmut, line 9: so I assume that you had to copy this class because we are not putting in eva-server a test dependency for eva-lib,  right? that could be done, but I'm ok duplicating this class if the test dependency is not straightforward.

sundarvenkata-EBI: yes, I didn't bother setting up a test dependency just for the sake of this one class which might be removed anyway if we eventually upgrade to nosql-unit 1.0.

PR #145, opened 2021-03-03
EVA-2327 - Migrate to cloudsmith artifactory

sundarvenkata-EBI:


sundarvenkata-EBI:
Github actions script [ran successfully](https://github.com/sundarvenkata-EBI/eva-ws/actions/runs/618996191).

dgva-server/pom.xml
```diff
@@ -82,7 +82,7 @@
 
     <profiles>
         <profile>
-            <id>Non-Travis</id>
+            <id>Non-GithubActions</id>
```
andresfsilva, line 85: I am wondering why here we are declaring a `Travis` profile which is empty (although used from the .travis.yml) and another `Non-Travis` profile which I don't see we are using. I know this code is old but if that is not being used it would be good to remove it

sundarvenkata-EBI: I think the Non-Travis profile is the one being used by default (see activeByDefault tag further below) so it doesn't have to be explicitly called out when it is being used. I guess the other profile is empty to explicitly say that the dependencies used by the default profile should not be used.
that is my understanding but I didn't want to create new problems taking those profiles out :smile: I have created [another ticket](https://www.ebi.ac.uk/panda/jira/browse/EVA-2349) to more deeply examine the need for these.

andresfsilva: That makes sense, thanks Sundar

andresfsilva (approved):

PR #149, opened 2021-07-27
EVA-2485: Prototype service for reading and storing counts  for ingestion pipelines

nitin-ebi:


count-stats/.gitignore
```diff
@@ -0,0 +1,33 @@
+HELP.md
```
andresfsilva, line 1: There is a `.gitignore` at the project level that is applied to the modules. Please remove this file and use the more general one

count-stats/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-resources-plugin</artifactId>
+                <version>2.6</version>
+                <configuration>
+                    <encoding>UTF-8</encoding>
+                    <delimiters>
+                        <!-- Provide a delimiter that doesn't clash with Spring ${} placeholders nor @ in URLs -->
+                        <delimiter>|</delimiter>
```
andresfsilva, line 89: There is some configuration or dependencies already in the pom that do not need to be included in the module pom. It is better not to duplicate them unless needed for specific reasons

nitin-ebi: removed

count-stats/.gitignore
nitin-ebi: removed

count-stats/src/main/java/uk/ac/ebi/eva/countstats/model/Count.java
```diff
@@ -0,0 +1,37 @@
...
+    private Long id;
+    @NotNull
+    private String process;
+    @NotNull
+    @Type(type = "StringJsonObject")
+    private String identifier;
+    @NotNull
+    private String metric;
+    @NotNull
+    private long count;
```
sundarvenkata-EBI, line 36: Sorry for the late request but I was wondering if we can also add a timestamp for easy tracking...

count-stats/src/test/java/uk/ac/ebi/eva/countstats/integration/CountStatsIntegrationTest.java
```diff
@@ -0,0 +1,83 @@
...
+                    applicationContext,
+                    "spring.datasource.url=" + postgreDBContainer.getJdbcUrl(),
+                    "spring.datasource.username=" + postgreDBContainer.getUsername(),
+                    "spring.datasource.password=" + postgreDBContainer.getPassword()
+            );
+        }
+    }
+
+    @Test
+    public void testSaveCount() throws Exception {
```
sundarvenkata-EBI, line 56: Is there a way to clean up database contents after every tests so that if we add other tests later they don't interfere with each other?

sundarvenkata-EBI (approved):

count-stats/src/main/java/uk/ac/ebi/eva/countstats/model/Count.java
nitin-ebi: added

count-stats/src/test/java/uk/ac/ebi/eva/countstats/integration/CountStatsIntegrationTest.java
nitin-ebi: updated

PR #158, opened 2022-02-07
EVA-2750 Add split and remapped fields to eva release stats API

andresfsilva:


eva-release/src/main/java/uk/ac/ebi/eva/release/dto/ReleaseStatsPerSpeciesDto.java
```diff
@@ -32,22 +32,32 @@
 
     private Long currentRs;
 
+    private Long remappedCurrentRs;
```
tcezard, line 35: I'm not sure what the meaning of remapped RS actually is. 
Now that an RS can be present in multiple assembly maybe its status in a species is less clear. 
An RS id can be `current` in one assembly and `remapped` + `split` in another.

eva-release/src/main/java/uk/ac/ebi/eva/release/dto/ReleaseStatsPerSpeciesDto.java
```diff
@@ -32,22 +32,32 @@
...
 
+    private Long splitRs;
+
     private Long deprecatedRs;
 
     private Long mergedDeprecatedRs;
 
     private Long unmappedRs;
 
+    private Long ssClustered;
```
tcezard, line 49: I'm not sure we'll have this value at the species level.
If it is a column in the database I'm not sure we want to expose it.

eva-release/src/main/java/uk/ac/ebi/eva/release/dto/ReleaseStatsPerSpeciesDto.java
apriltuesday: If I'm interpreting the [count gathering script](https://github.com/EBIvariation/eva-tasks/blob/master/tasks/eva_2750/gather_clustering_counts_from_mongo.py#L32) correctly, I guess this would be RS that were current/active in a previous assembly and have been remapped?

I agree that some of these numbers are a bit confusing per species, we should probably agree whether the species counts will sum over all assemblies for a taxonomy (in which case RS will be double-counted) or I guess only show counts for the target assembly of that taxonomy.

andresfsilva:
Agreed offline that the new columns do not apply to the "per species" stats and will be dropped

PR #179, opened 2024-09-20
EVA-3605 Filter out null assembly code

nitin-ebi:


apriltuesday (approved):

tcezard (approved):
Code makes sense but we should also check why there are `null` in the assembly codes.

PR #181, opened 2025-03-19
EVA-3770: Update exclusion filters for compatibility with Mongo 6

apriltuesday:


eva-server/src/main/java/uk/ac/ebi/eva/server/ws/RegionWSServer.java
```diff
@@ -126,6 +126,11 @@ public QueryResponse getVariantsByRegion(@PathVariable("regionId") String region
                 }
                 excludeMapped.add(docPath);
             }
+            // Recent versions of MongoDB do not allow projecting an embedded document with any of its fields.
+            // See https://www.mongodb.com/docs/v4.4/release-notes/4.4-compatibility/#path-collision-restrictions
```
tcezard, line 130: Do I understand this correctly that having both exclusion (projection) by a subdocument and its fields is not valid anymore. 
Would that apply to all fields like `annot` and `annot.xrefs` ?

apriltuesday: That's correct, however in reality the API only accepts values in [this map](https://github.com/EBIvariation/eva-ws/pull/181/files#diff-59efba96078b551afffce8f88f8af383563a5a28d77350340c0d8f2c4adf652eL80) so we technically only have to handle `files` and `files.attr` (called `sourceEntries` and `sourceEntries.attributes` in the response).

Compare these API calls:
* https://www.ebi.ac.uk/eva/webservices/rest/v1/segments/11%3A128446-128446/variants?species=hsapiens_grch37&exclude=sourceEntries&exclude=sourceEntries.attributes
* https://www.ebi.ac.uk/eva/webservices/rest/v1/segments/11%3A128446-128446/variants?species=hsapiens_grch37&exclude=annotation&exclude=annotation.xrefs

It's a strange undocumented behavior that I'm sure no users are using, but I'm guessing we use it internally.

tcezard: Explanation makes sense. 
Not commenting on the API behaviour ...

tcezard (approved):

nitin-ebi (approved):

PR #182, opened 2025-05-02
EVA-3644 Improve robustness of study browser

nitin-ebi:


eva-lib/src/main/java/uk/ac/ebi/eva/lib/metadata/eva/StudyEvaproDBAdaptor.java
```diff
@@ -43,11 +46,15 @@ public class StudyEvaproDBAdaptor implements StudyDBAdaptor {
...
         List<VariantStudy> variantstudies = new ArrayList<>();
-        for (EvaStudyBrowser dgvaStudy : dgvaStudies) {
-            if (dgvaStudy != null) {
-                variantstudies.add(dgvaStudy.generateVariantStudy());
+        for (EvaStudyBrowser evaStudy : evaStudies) {
+            if (evaStudy != null) {
+                try {
+                    variantstudies.add(evaStudy.generateVariantStudy());
+                } catch (Exception ex) {
+                    logger.error("Exception while generating variant study for {}. Exception {}", evaStudy, ex);
```
tcezard, line 56: I think that's great but with the amount of erro in the web-services, these will be difficult to detect. 
Maybe we should have an EVA specific string we can grep for to catch all data based errors.

tcezard (approved):

apriltuesday (approved):

eva-lib/src/main/java/uk/ac/ebi/eva/lib/metadata/eva/StudyEvaproDBAdaptor.java
nitin-ebi: updated

PR #208, opened 2026-07-24
EVA-4228 upgrade to kubernetes

nitin-ebi:


apriltuesday (approved):

eva-server/local-config/application.properties
```diff
@@ -0,0 +1,48 @@
...
+spring.data.mongodb.password=mongo_pass
+spring.data.mongodb.read-preference=primaryPreferred
+
+db.collection-names.files=files_2_0
+db.collection-names.variants=variants_2_0
+db.collection-names.annotation-metadata=annotationMetadata_2_0
+db.collection-names.features=features_2_0
+db.collection-names.annotations=annotations_2_0
+
+contig-alias.url=https://localhost:8080/contig-alias
```
tcezard, line 34: I don't think its a real problem but this endpoint won't exist in the docker compose.

eva-server/pom.xml
```diff
@@ -18,8 +18,7 @@
         </dependency>
         <dependency>
             <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter-tomcat</artifactId>
-            <scope>provided</scope>
+            <artifactId>spring-boot-starter-actuator</artifactId>
```
tcezard, line 21: spring-boot-starter-actuator is already defined below.

tcezard (approved):

eva-server/pom.xml
nitin-ebi: removed

eva-server/local-config/application.properties
nitin-ebi: yes, you are right. I just wanted a value for local so that it does not fail to start

